PR #7, opened 2018-03-26
US12756: Pagination

lukemoderwell:
_PR does:_
Pagination Stuff
- enables pagination 🎉 
- sets a max 16 posts to each page
- basic pagination styling
- removes the featured post styling (due to conflicts with pagination)

Misc fixes 
- links author names on topic pages
- creates absolutely positioned card blocks (per Adam George's feedback)


collections/_topics/money.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: topic
+name: money
+permalink: /topics/money/
+image: https://images.unsplash.com/photo-1500496733680-167c3db69389?ixlib=rb-0.3.5&s=0bb120921b4af2e0dcd01aa082aee04a&auto=format&fit=crop&w=1239&q=80
```
tcmacdonald: @lukemoderwell This is going to be tough to populate– we might need to come up with another solution for topic images.

_includes/_pagination.html
```diff
@@ -0,0 +1,30 @@
+{% if paginator.total_pages > 1 %}
+<div class="pagination text-center" style="width: 100%;">
+  {% if paginator.previous_page %}
+    <a href="{{ paginator.previous_page_path }}" class="previous">Previous</a>
+  {% else %}
+    <span style="color:#979797">Previous</span>
```
tcmacdonald: Inline styles?

collections/_topics/money.md
lukemoderwell: I'll scrap these for now until we come up with a good way to get images.

seancdavis: Looks like this is still hanging around.

_scss/_base.scss
```diff
@@ -2,6 +2,13 @@
   flex-wrap: wrap;
 }
 
+.cd_subheading {
```
seancdavis: Is there a reason you went with an underscore instead of hyphen?

_config.yml
```diff
@@ -12,20 +19,20 @@ plugins:
   - jekyll-feed
   - jekyll-redirect-from
   - jekyll-hubspot
+  - jekyll-paginate-v2
+
+pagination:
+  enabled: true
+  debug: false
+  collection: 'posts'
+  per_page: 8
```
seancdavis: I'm guessing this is just for demonstration purposes since we don't have 16 posts yet?

_scss/_base.scss
lukemoderwell: bad old `BEM` habit. Good catch

_config.yml
lukemoderwell: Right. It was more for testing.

collections/_topics/money.md
lukemoderwell: I left that and changed the front end to show a default image for all topics until we get that sorted out.

_scss/_base.scss
tcmacdonald: We have some loose direction [here](https://design.crossroads.net/developers/getting-started/coding-style) but nothing super specific. I'm guessing @seancdavis is speaking to general convention in `crds-styles`.

PR #9, opened 2018-03-28
Blog Categories

tcmacdonald:
This PR implements "blogs" as top-level categories. NOTE– I haven't set up the actual collections yet, just pulling in the data. 

Also worth noting, this update also adds support for by-passing Hubspot via environment variables. If you set `JEKYLL_ENV=development` it will skip the remote requests and rely solely on local data files.


tcmacdonald:
Closing because #11

PR #11, opened 2018-03-29
Blog Categories

tcmacdonald:
This PR implements Hubspot "blogs" as categories ([for example](https://deploy-preview-11--crds-mediaint.netlify.com/topics/)).

Also worth noting, this update adds support for by-passing Hubspot via environment variables. If you set `JEKYLL_ENV=development` it will skip the remote requests and rely solely on local data files.


_includes/_article-header.html
```diff
@@ -8,7 +8,7 @@ <h5 class="article_meta metadata flush">
       </h5>
       <h2 class="article_title section-header flush-top push-half-bottom">{{ page.title | escape }}</h2>
       <div>
-        <a href="{{ author.permalink }}">{{ page.author }}</a> |
+        <a href="{{ author.url }}">{{ page.author }}</a> |
```
lukemoderwell: was anything wrong with using `permalink`?

PR #16, opened 2018-03-30
US12825 - Tag Names and Search

seancdavis:
Add support for tag names which link to int search.

Depends on ample/hubspot-ruby#1.


_config.yml
```diff
@@ -3,6 +3,7 @@ email: info@crossroads.net
 description: No matter what your beliefs, you are welcome here.
 baseurl: "" # the subpath of your site, e.g. /blog
 url: "https://www.crossroads.net" # the base hostname & protocol for your site, e.g. http://example.com
+tag_search_url: "https://int.crossroads.net"
```
tcmacdonald: This could be a problem.

PR #27, opened 2018-04-05
Defect/de5211 update favicon on media.cr.net

galiotoweb:
Updates favicon to be the same as CRDS-Maestro and moves files to root as recommended


galiotoweb:
Passes Favicon Checker at https://realfavicongenerator.net/favicon_checker

PR #33, opened 2018-04-12
DE5288 - Author's pic on small screens

galiotoweb:
puts img min-width back in


_assets/stylesheets/_author.scss
```diff
@@ -5,6 +5,10 @@
     align-items: center;
   }
 
+  & .img-size-5 {
+    min-width: 5rem;
```
tcmacdonald: @galiotoweb Does this belong in `crds-styles`?

galiotoweb:
Talked to @zarahzachz, and realized instead of adding a CSS class, it would be better to match the media-object implementation in the DDK (a background image instead of an inline image).

PR #45, opened 2018-04-20
US12887 Podcasts Landing Page

galiotoweb:


_assets/stylesheets/_podcast.scss
```diff
@@ -0,0 +1,15 @@
+.media--podcast {
```
seancdavis: I know we see the use of double hyphens in a few places, but we've talked about trying to avoid that and sticking with one hyphen permutations to be more in-line with how Bootstrap handles naming.

_assets/stylesheets/_variables.scss
```diff
@@ -0,0 +1,18 @@
+//
```
seancdavis: I'd like to think about bringing in crds-styles as SCSS so we don't have to keep adding variables, but this is okay for now.

_assets/stylesheets/_base.scss
```diff
@@ -1,10 +1,20 @@
+.landing-header {
```
seancdavis: I see you didn't create this file, but naming it `base` doesn't make sense to me—it seems like there's a lot of odds and ends in there.

At the very least, I think you should move out home-specific styles into a home partial.

_includes/_podcast-card.html
```diff
@@ -0,0 +1,22 @@
+<div class="container">
```
seancdavis: Why does this file exist? I don't see it being used anywhere.

podcasts.md
```diff
@@ -0,0 +1,27 @@
...
+layout: default
+---
+
+<div class="container media--podcast">
+  <p class="flush-bottom"><a href="{{ site.url }}">Media</a>/</p>
+  <h2 class="section-header landing-header flush-top">Podcasts</h2>
+  <div data-card-deck class="card-deck">
+    <div class="cards-2x">
+      <div class="row">
+      {% for podcast in site.podcasts %}
```
seancdavis: This is where we'll add the partial, but I don't think that needs to be part of this story. I'd say you can leave the loop here as-is.

_assets/stylesheets/_podcast.scss
galiotoweb: Okay. I didn't hear that conversation. Happy to change it

_assets/stylesheets/_variables.scss
galiotoweb: I moved this to it's own file, because we were starting to get variables and notes to make changes throughout the CSS, and this way, we can delete just the one file once we incorporate crds-styles variables

_includes/_podcast-card.html
galiotoweb: Only because I thought we'd eventually need it. I can delete it

_assets/stylesheets/_base.scss
galiotoweb: Renamed _base.scss to _global.scss, and moved article-specific styles to it's own file

zarahzachz:
@galiotoweb A few things I noticed:
- The breadcrumb link isn't very clickable. One solution would be to give the breadcrumb `margin-bottom: -.5rem;` and `position: relative;` styles
- Add a space between the forward slash and the link
- The color of the forward slash should be `#979797` (as should be the card body - but since cards are getting refactored #YOLO)
- The space between breadcrumbs and the shared header is a little tight. Zepplin has it at 36px.

zarahzachz:
`#FF0000`

galiotoweb:
@zarahzachz All requested changes are done.

PR #65, opened 2018-04-27
US13262 - Placeholder Images

seancdavis:
Load images as very small and replace with actual size images after the DOM is ready


_includes/_featured_card.html
```diff
@@ -1,17 +0,0 @@
-<div class="card">
```
seancdavis: This is unrelated to this task, but I couldn't see where it was being used so I removed it.

tcmacdonald: I believe this was used initially to render one article super-big at the top of the homepage. That got removed somewhere along the line, so this is a good choice IMO.

_config.yml
```diff
@@ -161,4 +161,6 @@ assets:
 
 imgix:
   find: images.ctfassets.net/y3a9myzsdjan
-  replace: crds-media.imgix.net
\ No newline at end of file
+  replace: crds-media.imgix.net
```
tcmacdonald: Didn't we decide these values are irrelevant (i.e. overridden by the ENV plugin) and should be removed from `_config.yml`?

seancdavis:
@tcmacdonald: Made a couple opinionated changes here:

1. Explicitly state which images should be processed by using `data-optimize-img` attribute. Otherwise, we're going to loop through the header avatar and any images within the body.

2. I don't start trying to download the images until the header is rendered. This lets the header do its thing first and speeds up time to be interactive with the page. This result is awesome:

screen shot 2018-05-04 at 8 30 47 am

That is on par with the best of the best from some of my benchmark testing.

---

That being said, I have one outlying concern with this PR and that is that it does not address images within the body (or description 😉) of a page. I don't have a great solution for that.

We could look to process the html with the jekyll-contentful plugin and adjust the image sources at that point. Or we could educate editors to not use images within the body (given this being a media site and the content architecture being prescribed, that doesn't feel like an unreasonable request).

PR #67, opened 2018-04-30
DE5336: Pagination URL Issue

lukemoderwell:


_includes/_pagination.html
```diff
@@ -15,13 +15,13 @@
...
   {% else %}
-    <li><a href="/">1</a></li>
+    <li><a href="/articles">1</a></li>
   {% endif %}
   {% for count in (2..paginator.total_pages) %}
     {% if count == paginator.page %}
       <li class="active"><a href="#">{{count}}</a></li>
     {% else %}
-      <li><a href="/page/{{count}}">{{count}}</a></li>
+      <li><a href="/articles/page/{{count}}">{{count}}</a></li>
```
tcmacdonald: Obviously this solves the defect – but as this approach specific to articles, this solution won't work when pagination is implemented for other content types. 

Can you refactor this PR so we can pass in the name of the collection when including the partial? Perhaps something that looks like this...

```{% include _pagination.html collection=articles %}```

lukemoderwell:
@tcmacdonald I updated the include to have a var for customizing the URL. This solves sort term but we will still need a feature/story for implementing pagination across all content types. This will require changing the way our directories are setup and maybe more. We may want to spike on whether using paginate-v2 or some other solution is best.

PR #70, opened 2018-05-02
US13265 Header & Footer @ Buildtime

tcmacdonald:
# Don't merge yet

This work should not be merged until crdschurch/crds-styles#309 is resolved.


Gemfile
```diff
@@ -1,19 +1,19 @@
...
 gem "jekyll", "~> 3.7.2"
+gem 'activesupport'
 
 group :development do
   gem "pry"
 end
 
 group :jekyll_plugins do
-  # gem "jekyll-contentful", "~> 0.0.1", path: File.join(File.dirname(__FILE__), "../jekyll-contentful")
+  gem "jekyll-crds", "~> 0.0.1", path: File.join(File.dirname(__FILE__), "./vendor/gems/jekyll-crds")
```
seancdavis: Another way to write this is `File.expand_path('vendor/gems/jekyll-crds', __dir__)`. I tend to use that only because I don't have to use `File` twice, which is completely arbitrary.

PR #75, opened 2018-05-07
DE5393 - Preview Images

seancdavis:
Fixes:

- Preview hero image
- Preview author bio image

Other changes/additions:

- Combines scripts for preview into its own (deferred) manifest file
- Add frontmatter support for which javascript manifest file to load (`js_manifest`)
- Add frontmatter support for javascript variables. This will make any values from ENV available as global JS variables for the names passed to `js_vars`
- Adds frontmatter support for preloader. Add `preloader: true` will add the preloader to the view
- Rename `site.env` to `site.jekyll_env` so we can load `ENV` as a hash into `site.env` without being confusing. I felt like this more closely aligned with how node works with `process.env`.

The new frontmatter support is documented in the readme.


_layouts/default.html
```diff
@@ -50,6 +50,10 @@
     };
 
     var JEKYLL_ENV = '{{ site.env }}';
+    var CONTENTFUL_SPACE_ID = '{{ site.ENV["CONTENTFUL_SPACE_ID"] }}';
+    var CONTENTFUL_PREVIEW_TOKEN = '{{ site.ENV["CONTENTFUL_PREVIEW_TOKEN"] }}';
+    var IMGIX_DOMAIN = '{{ site.ENV["IMGIX_DOMAIN"] }}';
+    var IMGIX_SRC = '{{ site.ENV["IMGIX_SRC"] }}';
```
tcmacdonald: This is nasty. Shouldn't we have a `_layouts/preview.html` file so we can not expose this stuff to the other 99% of users?

_config.yml
```diff
@@ -129,6 +129,10 @@ contentful:
           draft: false
 
 defaults:
+  - scope:
+      path: ""
+    values:
+      js_manifest: application_deferred
```
tcmacdonald: I need you to explain this to me @seancdavis .

_includes/_js_vars.html
```diff
@@ -0,0 +1,5 @@
+<script data-js-vars>
+  {% for var_name in page.js_vars %}
```
tcmacdonald: This is clever.

PR #78, opened 2018-05-08
Feature/us12885 podcast details

galiotoweb:
Podcast Episode page template


_assets/stylesheets/_podcast.scss
```diff
@@ -192,11 +213,15 @@
       }
     }
 
-    .push-top-large {
+    .push-top-60 {
```
tcmacdonald: Should this eventually be refactored back into `crds-styles`?

_assets/stylesheets/_podcast.scss
```diff
@@ -278,4 +348,16 @@
       }
     }
   }
+
+  .push-top-60 {
```
tcmacdonald: This looks redundant. I see the same selector above...

_includes/_explore-podcasts.html
```diff
@@ -0,0 +1,27 @@
+<!-- Podcasts -->
+<div class="push-bottom row">
```
tcmacdonald: General FYI on using includes... I generally recommend not creating separate files unless they're used in more than one place.

_assets/stylesheets/_podcast.scss
seancdavis: My two cents is that yes, this belongs in crds-styles, but we really need to think hard about the approach:

- You're using `px` instead of `rem` here, where `rem` is the preferred method with positioning. 
- `60px` and `120px` not being divisible by `16px` (the base) feels arbitrary.
- It doesn't follow the DDK's convention of `.push-[amount]-[direction]`.
- We'd talked about if we were to put this in the DDK, that we'd want to think about using it like [Bootstrap 4.x](https://getbootstrap.com/docs/4.1/utilities/spacing/), where the number is more of a _factor_ and not an explicit pixel value.

So maybe work it into DDK with a series of factors relating to rems? (e.g. `.push-4-top { margin-top: 4rem; }`)?

galiotoweb: Not sure if we're going to need this on multiple pages or if it's isolated to this template, so I wanted to keep it separate unless we needed it in multiple places.

_includes/_explore-podcasts.html
```diff
@@ -0,0 +1,27 @@
...
+  <div class="col-sm-8 col-sm-offset-2">
+    <div class="push-bottom">
+      <h2 class="podcast-section-header flush-top">More content to explore</h2>
+    </div>
+    <div data-card-deck class="card-deck">
+      <div class="cards-4x">
+        <div class="row">
+          {% for podcast in site.podcasts limit: 3 %}
```
seancdavis: Whereas this feels like it _should_ be using an include because it's duplicating podcast card content (or if it's not, I'd argue we should simplify the design and use the podcast card).

_assets/stylesheets/_podcast.scss
galiotoweb: One is nested under .media-podcast-show and the other is nested under .media-podcast-episode so they didn't bleed out into the rest of the project.

Happy to move them to global styles if you think that's a better approach

_assets/stylesheets/_podcast.scss
galiotoweb: Okay. I can include this in crds-styles, and the class formatting was a miss on my part: I will change it to `.push-[amount]-[direction]`.

- `60px` and `120px` were Rob's direction, and those specific amounts were everywhere which is why I used them but I can convert them to rem round up or down, because otherwise those numbers are not very clear. In the long term, I see this as a problem, because designs are  not always going to have amounts of space neatly divisible by `16px`
- In Bootstrap 4.x, since those numbers are a _factor_ and not a pixel value, I think it's going to be confusing if we use it for sizes larger than 3rem. 
> Classes are built from a default Sass map ranging from .25rem to 3rem.

galiotoweb: - Discovered I wasn't using `.push-top-60` nor `.push-bottom-60` so they went the way of the Dodo. 
-`.soft-bottom-120` is now the brand new class `.soft-7-bottom` which may be found in global.scss

_assets/stylesheets/_podcast.scss
galiotoweb: Turned out I created this class and then didn't use it, so I deleted them all

PR #84, opened 2018-05-10
DE5414 - Podcast Image Size

seancdavis:
Make podcast images square on podcast list view.

Note:

- I'm still working on fading the full size images in, so it does feel a little jagged
- This puts a redirect in place for episodes => podcasts so /episodes doesn't throw a 404


tcmacdonald:
@seancdavis Sorry I'm just now seeing this. I took a very similar approach in c99e2ab8b62d067897277afb204ac0c77760cbfd which is already in `development`. 

I see a redirect in this PR... is that legit?

PR #85, opened 2018-05-10
DE5386 - Images

zarahzachz:
Add stitcher image to CR's S3 and point references to that url.


tcmacdonald:
Thanks for looking into this @zarahzachz. 

This PR tells me I was mistaken when I thought we had some references to files on the wrong Cloudfront distribution. My assumption was based on [this feedback from Google PageSpeed](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fmediaint.crossroads.net%2F)...

screenshot 2018-05-10 20 47 43

Can you please track down the URLs in the above screenshot (ie. figure out what Cloudfront distro is associated with them) and then investigate changing the Cache-Control headers for these files on S3 so that PageSpeed will quit complaining?

zarahzachz:
A cache-control value of `max-age=[number]` can be added to each directory (if that's the direction we want to go). I need to know a) what the max-age value should be (1 year? never?) and b) if targeting individual image directories is the best practice?

PR #91, opened 2018-05-14
Feature/us13483 podcasts on authors pages

galiotoweb:
- Add podcasts and headings to author.html template
- Modifies logic to match other places where we've used assign instead of a for loop
- Add filter in case someone enters an authors name in all caps in Contentful


seancdavis:
@galiotoweb: Nice work, here! I changed `capitalize_all` to `titleize` to be more in line with [ActiveSupport's inflectors](https://apidock.com/rails/ActiveSupport/Inflector/titleize). Those inflectors aren't in use here (unfortunately), but it's a common library in Ruby (especially Rails), so it's accepted terminology.

PR #115, opened 2018-05-18
DE5455 BG Image Validations

tcmacdonald:
Add migration to require song's bg_image


_migrations/09-add-validations-to-songs.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function (migration) {
+  migration.editContentType('song').editField('bg_image').required(true);
```
lukemoderwell: We should add this for all bgImg dependent layouts. I think podcasts, episodes, albums could also get the required flag.

Also we may want to consider the article image required until we roll out some logic to serve another layout for content with no images.

PR #135, opened 2018-05-24
DE5484: Video Sorting

lukemoderwell:


videos.html
```diff
@@ -5,8 +5,9 @@
   videos:
     per: 8
 ---
-{% assign featured = site.videos | slice: 0, 2 %}
-{% assign unfeatured = site.videos | slice: 2, site.videos.size %}
+{% assign date_sorted = site.videos | reverse %}
```
galiotoweb: I know this works, but have you had a chance to dig into the sort option for the new pagination approach?
It will simplify this code a little.
https://github.com/crdschurch/crds-media#sorting-collections

videos.html
```diff
@@ -3,10 +3,11 @@
...
 paginate:
   videos:
-    per: 8
+    per: 10
+    sort: date desc
 ---
-{% assign featured = site.videos | slice: 0, 2 %}
-{% assign unfeatured = site.videos | slice: 2, site.videos.size %}
+{% assign featured = page.videos.docs | slice: 0, 2 %}
+{% assign unfeatured = page.videos.docs | slice: 2, page.videos.docs.size %}
```
tcmacdonald: @lukemoderwell Pagination now supports 'offsets` which will exclude the first `n` items from the pagination set. I think this is what you're trying to accomplish here. Take a spin through the following documentation and let me know if that might be a cleaner approach for what you're doing here.. 

https://github.com/crdschurch/crds-media#offsetting-collections

lukemoderwell:
@tcmacdonald videos were not being sorted by most recent. https://github.com/crdschurch/crds-media/pull/135/files#diff-d81ded785495bbd305b0be98a49c6e9fR7

videos.html
seancdavis: @lukemoderwell: What you're saying is correct from the markup we'd see on the page. However, the idea is to use the load more button to only grab those videos from subsequent pages. Therefore, we're willing to take that risk at this point.

In this solution, you'll end up with two floating cards on every other page, which we don't want.

[See how this approach is being used on the music page](https://github.com/crdschurch/crds-media/blob/development/music.html). That's the direction we're heading with pagination.

PR #138, opened 2018-05-24
U13611: Add Related Videos To Songs

lukemoderwell:


_layouts/song.html
```diff
@@ -25,8 +31,35 @@ <h1 class="section-header flush-top">{{ page.title }}</h1>
...
       </div>
     </div>
+    {% if videos.size > 0 %}
+    <div class="row text-left">
+      <div class="col-sm-10 col-sm-offset-1">
+          <h3 class="text-lowercase font-size-large font-family-base-light white-faded ">watch</h3>
+          <div data-card-deck class="card-deck">
+            <div class="cards-4x">
+              <div class="row">
+                {% for video in videos %}
```
seancdavis: Why not use the video card?

lukemoderwell:
@seancdavis:
1. Image hovering doesn't work on the other video cards either. Can we add another defect/story for this? Maybe it needs to be adjusted on all cards...
2. I omitted the subtitle here until we really understand how we are sorting these videos. Agreed it wasn't the best approach but it was something until we have a better answer.
3. Nice call on the video card. I added a "dark" theme for the include so it doesn't look weird over an image or dark-bg.

PR #157, opened 2018-05-29
DEXXXX Performance

tcmacdonald:
Performance, take 2.


tcmacdonald:
@seancdavis Yea. This is me noodling approaches to inline critical CSS. So far, its not gone well. Maybe we could pair on it... I feel like I'm missing something here.

seancdavis:
Closing but leaving branch in tact in case we want to come back to it in the future.

PR #158, opened 2018-05-29
US13531 - Add chord charts and stems to song detail page

galiotoweb:
In addition to adding chord charts and stems to Contentful, I'm deleting subtitle and audio file from song and subtitle from album, because these fields are not being used anywhere.


galiotoweb:
I ran the migration in int. You can find an example at /songs/fantastic-song-2

_layouts/song.html
```diff
@@ -110,17 +134,11 @@ <h3 class="component-header">About This Song</h3>
             <a href="/albums/{{ page.album_slug }}">
               {{ page.album_title }}
             </a>
-            {% for author_name in podcast.authors %}
-              {% assign author = site.authors | where:"name", author_name | first %}
-              {% if author %}
-                <a href="{{ author.url }}">{{ author.name | titleize }}</a> {% if forloop.last == false %} and {% endif %}
-              {% endif %}
-            {% endfor %}
```
tcmacdonald: 💯

PR #166, opened 2018-05-31
Feature/us13323 video nav styles

zarahzachz:
Implement video sidebar navigation. Should be dropdown on mobile and a link list on desktop.

Test on `/videos`
No corresponding PRs.


videos.html
```diff
@@ -11,15 +11,23 @@
...
        <li class="list-unstyled soft-half-left border-left-blue"><p class="font-size-smaller">Recent</p></li>
+<<<<<<< HEAD
+      </ul> -->
+    </aside>
+
+=======
       </ul>
     </div>
     
+>>>>>>> development
```
tcmacdonald: Conflicts ^^

videos.html
```diff
@@ -11,15 +11,17 @@
...
-    <div class="col-md-2 text-left">
-      <ul class="hard">
+    <aside class="col-md-2">
+      {% include _sidebar-navigation.html %}
+      <!-- <ul class="hard">
        <li class="list-unstyled soft-half-left border-left-blue"><p class="font-size-smaller">Recent</p></li>
-      </ul>
-    </div>
-    
+      </ul> -->
```
tcmacdonald: Do we need to leave the commented HTML intact? I'd prefer to remove this.

videos.html
```diff
@@ -11,15 +11,17 @@
 {% assign unfeatured = page.videos.docs %}
 
 <div class="container">
-  <p class="breadcrumb flush-bottom hard"><a href="/">Media</a> /</p>
+  <p class="breadcrumb flush-bottom"><a href="/">Media</a> /</p>
```
tcmacdonald: This seems unrelated to the task at hand. Is this a problem specific to `videos.html` or do we have similar issues in other instances of the breadcrumb?

zarahzachz: I reimplemented the `hard` class - I just had a hard time clicking the breadcrumb to get back to media and this fixed that, but it is unrelated to this task.

PR #168, opened 2018-06-01
Defect/de5456 article text sizing

zarahzachz:
`<h*>` tags in the DDK have different font faces, sizes and weights associated with them. I've streamlined them so they're all one font face, weight and text-decoration. Sizing will vary based on which tag is used (as expected).

These styles are scoped to a `.markdown` class, which needs to be applied to any area we expect users to input markdown on the CMS. I've added this where applicable.

No corresponding PRs.


zarahzachz:
@tcmacdonald I removed the migration script and `package-lock` file. Ready to review.

_layouts/podcast.html
```diff
@@ -39,7 +39,7 @@ <h1 class="component-header flush-top">{{ page.title }}</h1>
   <div class="container content">
     <div class="row">
       <div class="col-sm-8 col-sm-offset-2">
-        {{ page.content }}
+        <p class="markdown">{{ page.content }}</p>
```
tcmacdonald: You're wrapping `page.content` in a `<p>` tag. That won't work if `page.content` contains markup.

PR #169, opened 2018-06-01
DE5500 - links don't work in podcast description

galiotoweb:
The `strip_html` filter was removing all markdown including links from the podcast description. This can be tested on /podcasts/spirit-stories

I also removed the `markdownify` filter because it's redundant. We are already converting markdown to html sitewide, so we don't need add the filter to individual elements.

I left `strip_html` in places where we wouldn't want links or different styling, for example the podcast cards as seen on /podcasts


_layouts/episode.html
```diff
@@ -64,18 +64,18 @@ <h2 class="component-header mobile-push-half-top">This episode</h2>
           <!-- Tab panes -->
           <div class="tab-content">
             <div role="tabpanel" class="tab-pane soft-bottom active" id="description">
-              <p>{{ page.content | strip_html | markdownify }}</p>
+              <p>{{ page.content }}</p>
```
tcmacdonald: You've got `page.content` inside a `<p>` tag. This won't work if `page.content` is parsed HTML.

PR #170, opened 2018-06-04
US13610 - Add Collections to Videos

seancdavis:
This is a beefy one. I've added inline comments for additional context.

I'm happy with this solution with one exception—the lack of tests. The existing tests are great because it shows the current functionality has not been disturbed, but setting up for new tests became problematic and super complicated, so I have not pursued it. Happy to talk out approaches to trying to find _some_ opportunity to write tests.


_layouts/video_tag.html
```diff
@@ -0,0 +1,34 @@
+---
+paginate:
+  videos:
+    per: 8
+    where:
+      tags: :title
```
seancdavis: New syntax for filtering collections.

_layouts/video_tag.html
```diff
@@ -0,0 +1,34 @@
...
+          {% for video in page.videos.docs %}
+            {% include _video-card.html image_size="4x" %}
+          {% endfor %}
+        </div>
+        <div class="loading hide">
+          {% include _preloader.html %}
+        </div>
+      </div>
+      {% assign link_root = "videos/tags/" | append: page.slug %}
+      {% include _pagination.html collection="videos" remote=true %}
```
seancdavis: We can't interpolate within an `include` block. Therefore, I am building the `link_root` separately. The support for this is reflected above, in the `_pagination.html` file.

_plugins/video_tags_generator.rb
```diff
@@ -0,0 +1,49 @@
+require 'fileutils'
+
+module Jekyll
+  class VideoTagsGenerator < Generator
```
seancdavis: This generator gathers all tags being used by videos and pipes them to the `_layouts/video_tag.html`. 

I played around with building a collection instead, but ultimately this played better because it resulted in minimal changes to the paginator gem.

vendor/gems/paging-mister-hyde/lib/paging-mister-hyde/paginator.rb
```diff
@@ -45,11 +45,11 @@ def paginate(type, collection)
           base = @page.instance_variable_get('@base')
           dir = @page.instance_variable_get('@dir')
           path = File.basename(@page.instance_variable_get('@path'))
-          path_sans_ext = File.basename(@page.instance_variable_get('@path'), '.*')
+          path_sans_ext = @page.url.chomp('/index.html').chomp('/')
```
seancdavis: Without this change, the paginator did not support adding pages to a URL different from the filename. In other words, while `/videos/tags/tag-name` would house the first page of videos for the "tag-name" tag, the second page would be placed at `/video_tag/page/2`. This change enables the second page to exist at `/videos/tags/tag-name/page/2`

vendor/gems/paging-mister-hyde/lib/paging-mister-hyde/paginator.rb
```diff
@@ -79,12 +81,22 @@ def paginated_collection(type)
...
       end
 
-      def sorted_collection(type)
-        collection = @site.collections[type].docs
+      def sort_collection(type, collection)
         if @cfg.dig(type, 'sort')
           sort_by, sort_in = @cfg.dig(type, 'sort') ? @cfg.dig(type, 'sort').split(' ') : nil
-          collection.sort_by! { |d| d.data[sort_by] } if sort_by
-          collection.reverse! if sort_in == 'desc'
+          collection = collection.sort_by { |d| d.data[sort_by] } if sort_by
```
seancdavis: The destructive methods (`sort_by!`, `reverse!`) were causing issues because they were getting back to `@site.collections`, which had the potential to screw up sorting (and filtering, shown below).

vendor/gems/paging-mister-hyde/lib/paging-mister-hyde/paginator.rb
```diff
@@ -79,12 +81,22 @@ def paginated_collection(type)
...
+          collection = collection.reverse if sort_in == 'desc'
+        end
+        collection
+      end
+
+      def filter_collection(type, collection)
+        return collection unless cfg.dig(type, 'where')
+        cfg.dig(type, 'where').each do |attr, value|
+          value = page.data[value[1..-1]] if value.start_with?(':')
```
seancdavis: `:title` would look for the value of `title` within the page's data, while `title` would try to match the string "title" directly.

vendor/gems/paging-mister-hyde/lib/paging-mister-hyde/paginator.rb
```diff
@@ -79,12 +81,22 @@ def paginated_collection(type)
...
+          collection = collection.select do |obj|
+            obj.data[attr].is_a?(Array) ? obj.data[attr].include?(value) : obj.data[attr] == value
```
seancdavis: If the attribute on an entry returns an array, we only look for the value to be included, otherwise we want it to match exactly. Therefore, we can support string and array fields.

README.md
```diff
@@ -127,9 +127,73 @@ You can implement an async 'load more' interaction automatically by ensuring you
...
+Finally, filtering supports array fields and looks for only a single match. The paginator does this automatically. Let's say your articles have tags via a `tags` field. And the current tag is the title of the page rendering the collection. You could do something like this:
+
+```liquid
+---
+...
+paginate:
+  articles:
+    per: 8
+    where:
+      tag: :title
```
tcmacdonald: Should this read `tags` per the description above?

_includes/_pagination.html
```diff
@@ -3,6 +3,8 @@
 
   {% if include.link_root %}
     {% assign root = include.link_root %}
+  {% elsif link_root %}
+    {% assign root = link_root %}
```
tcmacdonald: This condition confuses me. Under what circumstances would `link_root` be defined as a global in this partial?

tcmacdonald: Nevermind. I see your comment below which explains this.

_layouts/video_tag.html
tcmacdonald: This is gross but I get that it _works_.

_migrations/20180530100214-add-tags-to-videos.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function (migration) {
+
+  const video = migration.editContentType('video');
+        video.createField('tags').name('Tags').type('Array').items({ type: 'Symbol' });
```
tcmacdonald: Should we consider rolling a content type for tags and then doing a has_many association so that we can ensure users are repurposing existing tags vs. recreating them all the time? I'm a little concerned this will lead to lots of inconsistencies... "group" vs. "groups" vs. "small group" vs. "group-a-rama"

_layouts/video_tag.html
seancdavis: I would love a better way to do this. Any ideas?

_migrations/20180530100214-add-tags-to-videos.js
seancdavis: I really like this idea, but given that it touches multiple content types I think we should do it as a separate story.

PR #174, opened 2018-06-04
DE5498 - Blue header image

zarahzachz:
Remove that blue from that header.

No corresponding PRs.


_assets/stylesheets/_podcast.scss
```diff
@@ -181,9 +181,7 @@
   position: relative;
 
   .episode-header-background {
-    background: $cr-cyan;
-
-    & .podcast-background-image {
+    .podcast-background-image {
```
tcmacdonald: Did you intend to change the selector for `& .podcast-background-image`?

zarahzachz: Yes - That `&` is extraneous and since I was modifying the styles for the podcast background I figured it was okay to straighten up.

PR #179, opened 2018-06-05
US13609 - Most Viewed Videos

seancdavis:
Create a page that acts like a tag page (in the same vein as #170) specifically for the most viewed videos.

I chose not to paginate on this page because it seems like the _most viewed_ filter shouldn't contain all videos. Instead, I'm limited it to the 16 most popular videos.

Adding pagination would be simple if you (oh might reviewer) disagree. But we may want to think about how we exclude items from the paginated collection (i.e. items with a `0` view count, or some other lowest threshold).

---

While this PR contains the logic for working with the `view_count` field, this is really all built around crdschurch/crds-lambda#7.


zarahzachz:
@seancdavis I checked out the Netlify link and spun it up locally and the Most Watched section has videos on it, I just don't know if it's pulling the most viewed ones in.

I checked Contentful and am not seeing an uneditable `view_count` field. Should this be a thing?

PR #180, opened 2018-06-06
DE5502 - Podcast Sort Order

galiotoweb:


index.html
```diff
@@ -52,7 +49,7 @@ <h2 class="push-half-bottom">Articles</h2>
   {% endif %}
 
   <!-- Podcasts -->
-  {% if page.podcasts.docs.size > 0 %}
+  {% if site.episodes.size > 0 %}
```
seancdavis: Is this a good idea? I know we already know the production data, but this is saying if there aren't any episodes we're not going to show any podcasts. But what if the only podcasts we have are podcasts without episodes? (I know that's not the case, but should be be accounting for it?)

Shouldn't we still just check for the presence of podcasts?

podcasts.html
```diff
@@ -17,11 +13,8 @@ <h1 class="font-family-condensed-extra text-uppercase flush-top push-bottom">Pod
   <div data-card-deck class="card-deck">
     <div class="cards-2x">
       <div class="row">
-      {% for podcast in page.podcasts.docs %}
-        {% include _podcast-card.html %}
-      {% endfor %}
+        {% include _podcasts.html limit=site.podcasts.size %}
```
seancdavis: Not sure what this limit accomplishes. Is it because you _need_ to have a limit for the podcasts partial to work?

_includes/_podcasts.html
```diff
@@ -0,0 +1,16 @@
...
+{% for podcast in site.podcasts %}
+  {% assign podcasts = podcasts | append: podcast.title | append: '|' %}
+{% endfor %}
+
+{% assign orderedpodcasts = podcasts | split: '|' | uniq %}
+
+{% for item in orderedpodcasts limit: include.limit %}
+  {% assign podcast = site.podcasts | where:"title", item | first %}
+  {% include _podcast-card.html %}
+{% endfor %}
```
seancdavis: This is clever. Liquid is dumb so it forces you to be clever, but still, clever. 🥇

index.html
galiotoweb: That's a valid argument. Yes.

PR #181, opened 2018-06-06
US13322 - Video detail page

galiotoweb:
No corresponding PRs


_assets/stylesheets/_utilities.scss
```diff
@@ -31,25 +31,12 @@
...
   color: $cr-gray-lighter;
 }
 
-@media only screen and (max-width: $screen-sm-min) {
-  .mobile-center-block {
-    margin-left: auto;
-    margin-right: auto;
-  }
+.vertical-align-sub {
+  vertical-align: sub;
```
seancdavis: Should this go into crds-styles?

_layouts/video.html
```diff
@@ -1,5 +1,129 @@
...
 layout: default
 ---
+{% assign date_format = site.crds-media.date_format | default: "%b %-d, %Y" %}
+{% if page.source_url contains 'watch' %}
+  {% assign video_id = page.source_url | split: 'https://www.youtube.com/watch?v=' %}
+{% elsif page.source_url contains 'embed' %}
+  {% assign video_id = page.source_url | split: 'https://www.youtube.com/embed/' %}
+{% elsif page.source_url contains 'youtu.be' %}
+  {% assign video_id = page.source_url | split: 'https://youtu.be/' %}
+{% endif %}
```
seancdavis: Can we use a filter to clean this up? Perhaps something like:

```liquid
{% assign video_id = page.source_url | youtube_id %}
```

And then write a youtube filter that handles all this logic.

_layouts/video.html
```diff
@@ -1,5 +1,129 @@
...
+        <li role="presentation">
+          <a href="#transcription" aria-controls="transcription" role="tab" data-toggle="tab">Transcription</a>
+        </li>
+        {% endif %}
+      </ul>
+
+      <!-- Tab panes -->
+      <div class="tab-content">
+        <div role="tabpanel" class="tab-pane soft-bottom active markdown" id="description">
+          {{ page.content | markdownify }}
```
seancdavis: Didn't we say we don't need this because it comes in as HTML already? (I don't remember.)

_layouts/video.html
```diff
@@ -1,5 +1,129 @@
...
+        </div>
+
+        {% if page.transcription %}
+        <div role="tabpanel" class="tab-pane soft-bottom markdown" id="transcription">
+          {{ page.transcription | markdownify }}
```
seancdavis: This I'm not sure about. Does the jekyll-contentful gem only parse the body?

_layouts/video.html
```diff
@@ -1,5 +1,129 @@
...
+    <div class="col-sm-8 col-sm-offset-2">
+      {% for author_name in page.authors %}
+        {% assign author = site.authors | where:"name", author_name | first %}
+        {% if author %}
+          {% if forloop.first %}<p class="flush-bottom">Featuring {% endif %}
+            <a href="{{ author.url }}">{{ author.name | titleize }}</a> {% if forloop.last == false %} and {% endif %} {% if forloop.last %}</p>{% endif%}
+        {% endif %}
+      {% endfor%}
+
+      <p>{{ page.date | date: date_format }} {% if page.video_duration != nil %}
```
seancdavis: Do you think we could write another filter for dates that would prevent us from needing to assign a date_format on any page where we need it? That's become a lot of duplicated code.

In other words, could we simply do this:

```liquid
{{ page.date | format_date }}
```

Without needing to assign `format_date` on the page.

_layouts/video.html
```diff
@@ -1,5 +1,129 @@
...
+      <div class="col-sm-8 col-sm-offset-2">
+
+        <div class="push-bottom">
+          <h2 class="component-header flush-top">Recent Videos</h2>
+        </div>
+
+        <div class="card-deck carousel" data-crds-carousel="mobile-scroll">
+          <div class="feature-cards">
+            {% assign limited = site.videos | sort: 'date' | reverse | slice: 0, 10 | exclude: page.title %}
+            {% for video in limited limit: 9 %}
```
seancdavis: You've already sliced the videos, so why limit here?

_assets/stylesheets/_utilities.scss
galiotoweb: Yes. Potentially. We already have similar classes. Should I do that as part of this story?

_layouts/video.html
galiotoweb: We did. I did, but it wasn't working on this page without it. :(

_layouts/video.html
galiotoweb: I'm slicing at 10, but I only want 9. The slice is at 10 because if the first 9 includes the video on this current page, I don't want to display it under more content, and I don't want to display only 8 items because I only sliced the first 9.

_layouts/video.html
galiotoweb: For some reason, it was not working for this page. I'm happy to try and modify it

_layouts/video.html
galiotoweb: Yes. That sounds great

PR #189, opened 2018-06-14
Added Data automation ids

jasoncallibrity:
Looks like everything else I need to target will be pretty straight forward

Goal is to be able to target some of the functionality we have build in such as "See All" or the "breadcrumbs" and the "pagination".


zarahzachz:
@jasoncallibrity Will the naming convention of `*-landing` mess you up when automation is added to the actual media landing pages? I.e., you're using `article-landing` for the `index.html` page, but there is a `/article` landing page.

jasoncallibrity:
@zarahzachz They will be contained in their own PageObjects in our code so even if I reuse "article-landing" on both pages they are contained inside their own object. But you've given me pause on the name. Not sure that's the best name. Any suggestions? Naming is hard!

lukemoderwell:
@jasoncallibrity we may want to hold this PR until the new media landing goes live. It's going to change all of this code I.E. we won't have specialized containers for videos, articles, podcasts. Everything will be mixed together.

jasoncallibrity:
@lukemoderwell Awesome! Good to know. Is there an ETA on when those changes are going in? I'm kind of anxious to start doing some automation against a Netlify deployed site to see if it's more stable than our other deployments

PR #192, opened 2018-06-15
US13752 Series detail landing page

galiotoweb:
This has two corresponding PRs ~which need to be~ that were merged first:

https://github.com/crdschurch/crds-styles/pull/315
and https://github.com/crdschurch/crds-styleguide/pull/329

~Then I need to run `bundle update crds-styles` in the crds-media branch~

The migration PR needs to be merged to update the content model. Without the migration the related messages won't show up.
https://github.com/crdschurch/crds-contentful-migrations/pull/1

This PR creates the series detail page, updates the message content model, and cleans up some CSS


seancdavis:
@galiotoweb: Those PRs are now merged.

_assets/stylesheets/_utilities.scss
```diff
@@ -20,15 +20,12 @@
...
+.soft-7-bottom {
+  padding-bottom: 7rem;
+}
 
-    &:nth-child(odd) {
-      clear: left;
-    }
-  }
+.img-size-5 {
+  min-width: 5rem;
```
seancdavis: Thinking this name should reflect that it is a min width and not an explicit width.

_layouts/series.html
```diff
@@ -1,5 +1,70 @@
 ---
 layout: default
 ---
+<div class="media-series-detail">
+  <div class="jumbotron jumbotron-xl push-14-top push-2-bottom" style="background-image: url('{{ page.bg_image | imgix: site.imgix }}?{{ site.imgix_params.placeholder }}')" data-optimize-bg-img>
+    <div class="jumbotron-content">
+      <div class="container">
+        <div class="row">
+          <div class="col-sm-5 col-sm-push-7">
+            <img class="feature-img" src="{{ page.image | imgix: site.imgix }}?{{ site.imgix_params.placeholder_square }}" sizes="{{ site.image_sizes.podcast_thumb }}" data-optimize-img>
```
seancdavis: For this to work properly, it's going to need to be set with CSS to fill its container (100% width). Otherwise, you end up with this effect:

_assets/stylesheets/_utilities.scss
galiotoweb: I didn't create this. I just moved it from global.scss to utilities.scss

zarahzachz:
@galiotoweb Looks nice! Got a comment and a question:

Might want to use the `md` column classes for your layout. For series with long titles, this layout can happen. Increasing the breakpoint can fix this. Unless this was the design intention. Just pointing it out.
screen shot 2018-06-21 at 11 35 51 am

Also, it looks like your original work had the series artwork as squared images. This seems to have gone away since your latest work. Is this intentional?

PR #193, opened 2018-06-15
US13325: New Media Page

lukemoderwell:
PR covers:
- new layout
- adds `jumbocard` element
- adds `media-label` element
- `cards-2x` at larger screens and carousel at mobile

Still missing:
- changing the media label based on content type. This should be handled in [this story](https://rally1.rallydev.com/#/66096747656d/detail/userstory/228433790176)


_includes/_media-object-featured.html
```diff
@@ -0,0 +1,18 @@
+<!-- Tate: Use variables to toggle .featured in _media-object.html instead? -->
```
seancdavis: Can we remove?

seancdavis:
Looks really good. Nice work, @lukemoderwell!

PR #196, opened 2018-06-18
US13751 - Series landing

zarahzachz:
Update previous series section headers to have a smaller font size and include the word "series" after each year.


seancdavis:
@zarahzachz: I know the bulk of this has already been implemented, but the pixellated placeholder images weren't implemented correctly here. I believe the difference is you want to make the images 100% width so they stretch to the correct size even when they are super small.

screen shot 2018-06-18 at 3 05 04 pm

Also, you could use some padding after the last year. It's hugging up pretty close to the footer.

PR #197, opened 2018-06-18
US13610 - Add Collections to Videos

seancdavis:
Fixes an issue with load order of plugins by running the paginator from video tags builder.


seancdavis:
This also addresses a styling issue that would result in the sidebar nav floating above the footer when there weren't enough videos to push it down.

PR #207, opened 2018-06-21
US13828 - Labels

zarahzachz:
### Problem
There are local styles overriding `label`.

### Solution
Remove local styles and pull in utility classes to achieve the same look.

*Note*
I kept the mobile-specific styles because it felt unique to this layout.

**Please review after corresponding styles/styleguide PRs are merged.**


galiotoweb:
The corresponding styles/styleguide PRs are merged! 🎨

PR #213, opened 2018-06-27
DE5570: Set Explicit Img Widths

lukemoderwell:
add widths for media objects & podcasts


galiotoweb:
@lukemoderwell Here's some screenshots of what I'm talking about with the featured images on the landing page.

They jump to wider than they should be before jumping back to their original size:

PR #219, opened 2018-06-29
US13769: Updated Song Layout

lukemoderwell:


_assets/stylesheets/_external-music-links.scss
```diff
@@ -0,0 +1,13 @@
+.source-links  {
+  @media (max-width: $screen-sm) {
+    & {
```
seancdavis: Does this do anything? I thought it just meant the current context, but does it provide some utility?

_assets/stylesheets/_utilities.scss
```diff
@@ -28,6 +44,14 @@ a.hover-text-gray-lighter:hover * {
   font-size: 10px;
 }
 
+.font-size-giant {
+  font-size: 48px;
+  @media (min-width: $screen-sm) {
+    font-size: 90px;
+    line-height: .89;
```
seancdavis: Curious, why is this necessary?

_assets/stylesheets/_external-music-links.scss
lukemoderwell: @seancdavis yeah this must have been hanging around from something I tried. Need to clean it up.

_assets/stylesheets/_utilities.scss
lukemoderwell: Without the line-height the large headings were looking too spacious. The magic number comes from Zeplin.

lukemoderwell:
@seancdavis this should be good to go now. One exception might be the layout for the stream or buy on mobile. If they have 3 links things look a bit awkward still but I've move the 3rd item from center to the left. Because of the 2 column layout on small screens things get weird when you have an odd number of links.

PR #220, opened 2018-06-29
Defect/de5571 placeholder image ghosting

zarahzachz:
### Problem

Images in featured media section of media homepage "jump" on load. Also found that the placeholder background image repeated itself when it shouldn't.

Bonus problem: Image on the podcast episode header "flashed" while loading. 

### Solution

I removed the inline styles for the background images. Because these new jumbocards were essentially behaving like jumbotrons, I added our jumbotron background style rules to its CSS. This prevented the repeating imagery and cleaned up the markup considerably.

I also removed the padding solution from the jumbocard styles. Using padding to set the height was breaking our Imigix solution, which relies on a explicit height of it's parent to resize itself appropriately. To fix this, I added a series of image heights at various breakpoints and screen orientations.

Bonus solution: I increased the `z-index` of the podcast header image and now it doesn't flash when loading.

No corresponding PRs.

### DIFF
--- a/_assets/stylesheets/_jumbocard.scss
+++ b/_assets/stylesheets/_jumbocard.scss
@@ -1,10 +1,35 @@
 .jumbocard {
+  height: 275px;
   position: relative;
-  height: 0;
-  padding-top: 75%; // 4x3
 
-  @media (min-width: $screen-sm) {
-    padding-top: 56.25%; // 16x9
+  @media screen and (min-width: $screen-xs) and (orientation: portrait) {
+    height: 390px;
+  }
+
+  @media screen and (min-width: $screen-sm) and (orientation: portrait) {
+    height: 380px;
+  }
+
+  @media screen and (min-width: $screen-md) and (orientation: portrait) {
+    height: 400px;
+  }
+
+  @media screen and (min-width: $screen-lg) {
+    height: 420px;
+  }
+
+  @media screen and (min-width: $screen-xs) and (orientation: landscape) {
+    height: 300px;
+  }
+
+  @media screen and (min-width: $screen-md) and (orientation: landscape) {
+    height: 400px;
+  }
+
+  &[style*='background-image'] {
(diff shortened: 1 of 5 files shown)


_assets/stylesheets/_jumbocard.scss
```diff
@@ -1,10 +1,35 @@
 .jumbocard {
+  height: 275px;
```
seancdavis: Is there a way to fix the script rather than specifying explicit heights? This feels messy because there are so many rules and it is making an assumption on its width to achieve the dimension ratio we're after.

I'd rather see if you can fix this via the script rather than writing explicit style rules.

There's [a lot going on in the script](https://github.com/crdschurch/crds-media/blob/development/_assets/javascripts/components/images.js), but it's documented with comments, so hopefully you can make your way around quickly. If it's calculating height wrong because of padding, it may just be some math or a calculation we have to change. Maybe [this line](https://github.com/crdschurch/crds-media/blob/development/_assets/javascripts/components/images.js#L114) is causing a problem?

zarahzachz: @seancdavis I discussed this approach with Taylor. He said that there would be some upcoming refactor work for the image script due to performance enhancement opportunities and suggested I move forward with the quick solution now and document the image updates in a future story (it's US14095 on the backlog).

I can look at fixing the JS script if you think that's the better approach. Let me know what you recommend.

zarahzachz:
@seancdavis I looked into the JS and it turned out to be a simple CSS fix 😅. I applied `position: relative` to the parent element during the background image process and tested it throughout the site (song details page, articles page, podcast series, podcast episodes, and albums). 

The only weirdness I noticed was in how the images on the albums page load, but this lives in INT so it's not related to my fix. If you wanna see what's happening, go to `/albums/whatever-pleases-you.html`. I'll reach out to Jason to see if this is a known defect.

Ready to review!

PR #227, opened 2018-07-05
DE5593: Sorting Articles on Topic Page

lukemoderwell:
In addition to sorting articles, I made a small css tweak to cover a [recently merged PR](https://github.com/crdschurch/crds-media/pulls?q=is%3Apr+is%3Aclosed).

### DIFF
--- a/_layouts/topic.html
+++ b/_layouts/topic.html
@@ -5,7 +5,7 @@ layout: default
 <div class="container">
   <h1 class="section-header">{{ page.title }}</h1>
 
-  {% assign articles = site.articles | where:"topic", page.title %}
+  {% assign articles = site.articles | where:"topic", page.title | reverse %}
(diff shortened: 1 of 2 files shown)


_layouts/topic.html
```diff
@@ -5,7 +5,7 @@
 <div class="container">
   <h1 class="section-header">{{ page.title }}</h1>
 
-  {% assign articles = site.articles | where:"topic", page.title %}
+  {% assign articles = site.articles | where:"topic", page.title | reverse %}
```
seancdavis: This doesn't seem to have fixed the issue. In reading this, I wouldn't expect it to. You're reversing the collection, but are you certain of its order in the first place? Thinking you're going to want to look at ordering with `sort_by` first.

screen shot 2018-07-05 at 2 06 37 pm

PR #229, opened 2018-07-06
US13745 - Redirects

seancdavis:
/media doesn't keep the prefix when redirecting, so I've lopped off those prefixes. I don't anticipate this causing a conflict, but if we'd rather have Dillon keep the "/media" prefix in the redirect URL I can do that.


seancdavis:
@dillon-courts was able to complete a redirect that maintains the `/media` prefix, so we don't need this change.

PR #237, opened 2018-07-11
DE5618: Program Link Not Showing

lukemoderwell:
**Problem**: the program link (and audio link) weren't showing even if content was added

**Solution**: add markup to show them when present

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -251,6 +251,7 @@ contentful:
           authors: author/full_name
           image: image/url
           source_url: source_url
+          audio_source_url: audio_source_url
           program: program/url
           video_duration: video_duration
           topic: category/title
--- a/_config.yml
+++ b/_config.yml
@@ -252,7 +252,7 @@ contentful:
           image: image/url
           source_url: source_url
           audio_source_url: audio_source_url
-          program: program/url
+          program_url: program/url
           video_duration: video_duration
           topic: category/title
           topic_slug: category/slug
--- a/_layouts/video.html
+++ b/_layouts/video.html
@@ -17,38 +17,24 @@ layout: default
       {% endif %}
 
       <!-- Start breadcrumbs -->
-      <div class="row">
-        <div class="col-xs-8 col-lg-10">
-          {% assign url_array = page.url | split: '/' | shift | pop %}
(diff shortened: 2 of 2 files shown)


_layouts/video.html
```diff
@@ -17,24 +17,38 @@
...
+          {% capture url_string %} {{ page.url | replace: '/', " " }}{% endcapture %}
+          {% assign url="" %}
+          <ol class="breadcrumb push-top hard flush-bottom">
+            <li><a class="text-cyan" href="/">Media</a></li>
+
+            {% for breadcrumb in url_array %}
+            {% capture first_word %}{{ url_string | truncatewords: 1, "" }}{% endcapture %}
+            {% capture url %}{{ url }}/{{ first_word }}{% endcapture %}
+
+            <li><a class="text-cyan" href="{{ url }}">{{ breadcrumb | replace: '-', ' ' | capitalize }}</a></li>
```
zarahzachz: Why is `text-cyan` being added here?

_config.yml
```diff
@@ -251,6 +251,7 @@ contentful:
           authors: author/full_name
           image: image/url
           source_url: source_url
+          audio_source_url: audio_source_url
           program: program/url
```
zarahzachz: What do you think about changing `programing` to `program_url` or `program_pdf`? Something that's a little more semantic.

lukemoderwell: I usually just match whatever is setup in the migration but that makes sense. Would we also change the migration file?

_layouts/video.html
lukemoderwell: Based on this [design](https://projects.invisionapp.com/d/#/console/13922504/291615856/preview) it looked like they were cyan.

_config.yml
zarahzachz: No, you can change the green half of that line in the config file and leave the black half the way it is. This is just mapping the Contentful/migration side of things to our Jekyll site, so you could have it say something like `peanut_king: img/url` and wherever you reference `{{ page.peanut_king }}` you'd get the `img/url` data entered in Contentful. (It's a contrived example, but I stand by it!)

PR #240, opened 2018-07-12
Defect/de5621 description zero state

zarahzachz:
### Problem
If no content provided in tab interface for /episode, /video, or /song, the layout displayed an empty tab that looked gross.

### Solution
Doing a couple things for this. 

First, I added some logic to hide the tabs if no content has been added to any of them. I also hid the "description" tabs if `page.content` was empty.

Second, I made the tabs a bit more flexible by making the first tab, whether it's song lyrics or video transcription, the active tab.

### Bonus fix:
I removed the deprecated `.markdown` class.


_assets/javascripts/components/tabs.js
```diff
@@ -0,0 +1,3 @@
+$(document).ready(function() {
+  $('.nav-tabs a:first').tab('show');
```
tcmacdonald: Will this throw an error if there are no tabs on a page?

zarahzachz: I just checked it and while there is an error on the video details page, helpfully saying `undefined`, this wasn't caused by this tabs.js stuff. I checked mediaint and it's presently there.

My tabs script doesn't cause any additional errors if no tabs are present from what I can tell.

zarahzachz:
@tcmacdonald I refactored the tabs to live in their own partial and added logic to show/hide tabs based on whether there was content. The entire tab block will only be show if any tab field has content (so no tabs if there is no tab field content).

Ready for review.

_includes/_tabs.html
```diff
@@ -0,0 +1,17 @@
+<div>
```
seancdavis: This is a really clever solution.

_layouts/episode.html
```diff
@@ -34,40 +34,14 @@ <h1 class="font-family-condensed-extra font-size-h3 text-uppercase text-white fl
...
 
-            {% if page.show_notes %}
-            <div role="tabpanel" class="tab-pane soft-bottom" id="show-notes">
-              {{ page.show_notes }}
-            </div>
-            {% endif %}
+        {% assign tabs = 'content show_notes transcription' | split: ' ' %}
+        {% if page.show_notes or page.transcription or page.content != blank %}
+          {% include _tabs.html tabs=tabs %}
+        {% endif %}
```
seancdavis: One thought, just to take it a step further. What if you define the tab names in `_config.yml` so you don't have to use liquid. And then instead of including an if statement everywhere there might be tabs, you wrap your tabs partial in an if statement?

In other words, what if on this episode page you could have something like:

```liquid
{% include _tabs.html tabs=page.tabs %}
```

And the tab partial wraps itself in:

```liquid
{% if include.tabs.size > 0 %}
```

Or something like that. May make it even cleaner?

_layouts/song.html
```diff
@@ -89,56 +89,14 @@ <h6 class="card-title text-white push-quarter-bottom">{{ video.title }}</h6>
 <div class="container soft-7-bottom">
   <div class="row push-bottom soft-bottom">
     <div class="col-sm-8 col-sm-offset-2">
+      {% if page.details or page.lyrics or page.chords %}
```
seancdavis: Since we're on the topic of being DRY and clever, I was thinking there is a lot to remember if we want to add or remove a tab from a view. We have to know the tab set and then we repeat it in several cases.

If you implement my suggestion on defining the tab names for any given collection, perhaps you could write a simple generator that attaches some sort of method on each doc in the collection, say `has_tab_content` that would loop through the tab names and see if the doc has that value present.

_plugins/filters/pry_filter.rb
```diff
@@ -0,0 +1,11 @@
+require 'pry'
```
seancdavis: Also super clever. Nice!

zarahzachz:
Oh yeah, tab examples can be found here:
No tab content: `/videos/whats-crossroads`
One tab: `/videos/dude-this-is-a-really-nice-vidyo`
All tabs: `/videos/redemption-song-live`

I've only included video examples, but the behavior should be the same across videos, songs and podcast episodes.

_layouts/episode.html
zarahzachz: The reason I ended up going with this approach was because I couldn't get the frontmatter solution working. I agree that it would be a nicer approach, but adding a tabs section to the config file or template frontmatter resulted in `nil` being returned every time.

PR #243, opened 2018-07-16
Feature/us13756 update article ui

zarahzachz:
**To be merged *after* styles PRs are merged and `bundle update crds-styles` is run.**
Corresponds with crdschurch/crds-styleguide#334 and crdschurch/crds-styles#319

---
### Problem
Design for article details page is out-dated.

### Solution
Update styles so article details page matches new designs (https://app.zeplin.io/project/5acf6d62553bdae0669efdd2/screen/5b48b298ac910ac764fa76f2).

Created a migration in crdschurch/crds-contentful-migrations#8


_layouts/article.html
```diff
@@ -3,6 +3,88 @@
...
+            <div class="card-deck carousel" data-crds-carousel="mobile-scroll" data-card-deck>
+              <div class="feature-cards feature-cards-3x">
+                {% assign limited = site.articles | sort: 'date' | reverse | slice: 0, 4 | exclude: page.title %}
+                {% for post in limited limit: 3 %}
+                  {% include _article-card.html %}
+                {% endfor %}
+              </div>
+            </div>
+
+            <p><a href="/articles">Read more</a></p>
```
lukemoderwell: @zarahzachz we talked about this but I feel like there are 2 approaches that we could take going forward: 
1. Move this link to be flush with the 'recent articles' headline floated right and call it 'See All Articles' or something like that. 
2. Call this 'load more' and then just follow the same pattern we use elsewhere (fetching more article cards)

_assets/stylesheets/_cards.scss
```diff
@@ -67,18 +78,41 @@
...
-    }
-
     .card-block {
       .card-title:hover {
         text-decoration: underline;
       }
     }
   }
+
+  // styles for article-card.html
```
lukemoderwell: Should these be under an `.article-card` class similar to how you are doing the `.article-tpl` below? If I have a video card that has a `.title` should it inherit these styles?

zarahzachz: 🤔 You're probably right about this. I originally had these styles in `articles.scss` but moved them because the article-card is being included in a bunch of templates. It makes sense to add a class to the article-card partial to make sure we don't see regressions.

_layouts/article.html
zarahzachz: For now, I'll go with option 1 since Becky said the ask was to link to the /articles page.

zarahzachz:
Great feedback - I'm taking care of the following:
- [ ] Add `.article-card` namespace wrapper to `_article-card.html` include
- [ ] Change "Read more" link to "See all articles" and move to inline with header
- [ ] Remove `push-quarter-top` from `_article-card.html`

lukemoderwell:
@seancdavis totally agree your 2nd point. I've been shut down in the past for going with visual similarity vs actual px in Zeplin but I really think the `600` weight is much chunkier and I believe @amplerob would agree that matching `font-weight` visually would be preferable.

zarahzachz:
@seancdavis I agree with everything.

1. I think the grid solution is worth adding to the DDK, but should probably go in when we've settled on an updated card style. This feels like part of a new system and I want to see where we land on all the base pieces before bringing in the grid solution... if this makes sense.
2. I think Rob's probably using the official 600 weight font from Typekit, while we're using CSS to manipulate a 400 weight font. That could explain the visual discrepancy between his design and our work. Because the Typekit font we're using is a 400 weight, I went to that.
3. Thanks!
4. Good call - I've refactored the CSS to use DDK utility classes. As Brian's pointed out before, these aren't necessarily new components (just cobbled together images and headers). One thing I want to point out is the design is using a lot of sizes and line heights we don't have in the DDK. I went with utilities that most closely represent the design ask.

Ready to review.

PR #244, opened 2018-07-18
Feature/us14275 update article landing ui

zarahzachz:
### Problem
`/articles` should match new design [here](https://app.zeplin.io/project/5acf6d62553bdae0669efdd2/screen/5b48b30fc18fd03c7e9d709e).

### Solution
Had to create a new card style that behaves as block elements (card image and content are each block style elements) on mobile, and overlay elements on desktop (card image is behind the content).

Card headline needed to be truncated at 55 characters (Becky proposed 60, but space for the `...` needs to be accounted for).

Updated the pagination include so an optional "label" parameter could be passed in. I noticed the individual templates have differing CTA language.


_includes/_pagination.html
```diff
@@ -11,7 +11,7 @@
 
   {% if include.remote == true %}
   <div class="load-more">
-    <a href="#" data-role="load-more" data-paginate="{{ include.collection }}">Load more</a>
+    <a href="#" data-role="load-more" data-paginate="{{ include.collection }}">Load more {{ include.label }}</a>
```
seancdavis: Won't this leave an empty space at the end of the link if `include.label` is `nil`?

_assets/stylesheets/_cards.scss
```diff
@@ -91,3 +91,100 @@
     }
   }
 }
+
+
+
+
+.overlay-media-card {
```
seancdavis: Why did you create a new partial with new styles instead of using what's [being used on the home page](https://github.com/crdschurch/crds-media/blob/development/_includes/_jumbocard.html)? I believe Rob's intention was that these are the home page cards.

_includes/_pagination.html
zarahzachz: Yes. I considered this and figured a space at the end of an inline link is better than asking people to remember to add a space before the optional variable. I can change it back, but this felt like a nicer implementation.

_assets/stylesheets/_cards.scss
zarahzachz: The cards on the homepage don't behave the way Rob illustrated they should on mobile. What Rob's designs indicate is that the image and card text are stacked on mobile and move to an overlay-type position on desktop. I'm all for making the code reusable and can opt for the `jumbocard` strategy if that's preferred...

_includes/_pagination.html
seancdavis: @zarahzachz: What about something like `{{ include.label | prepend: ' ' }}`?

zarahzachz:
@seancdavis I updated the font weights and renamed the card class. It's still subject to change, but I feel like calling them `jumbocards` also implies how they should be used (like jumbotrons). Coming up with a name scheme might be easier once all the cards are designed/refactored...

I also talked to Becky updating the homepage with these new cards. She feels that should be a defect so we can move the article story through the board.

PR #245, opened 2018-07-18
US13768: Update Music Landing

lukemoderwell:
Updated music landing with now with 10000% moar `jumbocards`. Took the approach of limiting banner treatment to the 4 most recent songs. Everything else appears as a song card and uses the load more behavior.

_As per BMiller: gradient background have been removed._

[Related PR](https://github.com/crdschurch/crds-contentful-migrations/pull/9) to set up content model changes

### DIFF
--- a/_assets/stylesheets/_jumbocard.scss
+++ b/_assets/stylesheets/_jumbocard.scss
@@ -91,6 +91,7 @@
     }
 
     .jumbocard-title {
+      line-height: .91;
       @media (min-width: $screen-sm) {
         font-size: 90px;
         line-height: .89;
--- a/_assets/stylesheets/_jumbocard.scss
+++ b/_assets/stylesheets/_jumbocard.scss
@@ -76,13 +76,25 @@
   }
 }
 
-.jumbocard-music {
+.music-tpl .jumbocard {
   @media (min-width: $screen-sm) {
     padding-top: 40rem; // 640px
   }
 
   @media (min-width: $screen-md) {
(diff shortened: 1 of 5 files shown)


lukemoderwell:
@seancdavis thanks for the review. 
Your breadcrumb is missing "Music" from the design. I'm not sure if that was intentional or not.

Why offset 4 when the design shows 3?

Do you think we should have a heading above the smaller cards like "New Music" or something like that?

What do you think about implementing the new card styles for these songs rather than using the old one? Show the play icon and duration and updated heading?

I appreciate you removing the dark background per BM's direction. However, we've run into issues with this, in particular when there are songs that have an image, but not a background image. In those cases we should fallback to using the image but should have the dark gradient above it. But if the background image does exist, then don't show the gradient. I believe LeeAnne or Sarah have done this in other places.

call_to_action, featured_label, and featured_text are confusing to me. Specifically, featured_label and featured_text seem like really arbitrary terms and as an editor I wouldn't know which goes where. Perhaps the line below the title should bring back the old subtitle field. In that case, then featured_label may be a little easier to understand. call_to_action seems okay.

Eventually, we'll be able to move these fields off to another feature entry when we get there, so it won't be as big of a deal.

lukemoderwell:
@seancdavis added the following
- [x] used the `media card` with some slight tweaks to enable the `media-label`
- [x] liquid logic to use the gradiated background when the image is used but _not_ for the bg_image
- [x] config changes to support `featured_subtitle` instead of `featured_text`

PR #249, opened 2018-07-23
Defect/de5623 consistent template white space

zarahzachz:
### Problem
We're seeing inconsistent spacing problems from template to template - specifically around the footer element.

### Solution
I noticed in Rob's designs that the space between the content and footer varies between no space and 120px. I'm proposing we lean on Jekyll's layout feature a little more by creating two base templates: `default.html` wraps the page content in a div with `margin-bottom: 120px;` while `default_footer_flush.html` has no space between the footer and content. 

I went through the existing templates and removed a lot of the layout utility classes to give each template a more consistency.

I'm not sure if this is the *correct* solution, but I think if we lean on official layouts more, devs won't need to jerry-rig their own layout solutions on every page they work on - they can just plug in the appropriate layout and add the components that page needs. Willing to discuss a better solution.


_layouts/default_footer_flush.html
```diff
@@ -0,0 +1,15 @@
...
+      height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+
+    {% include header.html %}
+    {{ content }}
+    {% include footer.html %}
+    {% if page.preloader %}{% include _preloader.html %}{% endif %}
+
+    <script async type="text/javascript" src="{% asset {{page.js_manifest}} @path %}"></script>
+  </body>
+</html>
```
seancdavis: I like how this looks in that you still have to write the main wrapping tags—html, body—because there isn't any real logic there. But I think you can be more DRY and protect a little better against regressions. Specifically:

- The preview class only needs to be on the layout the preview page is using.
- Seems silly, but I would create a partial for GTM and footer script. I say this because then we only have to update these in one place if they ever change.

PR #256, opened 2018-07-24
Feature/us14402 article image gradient

zarahzachz:
### Problem
Header image on article details page doesn't have a gradient.

### Solution
Use `.bg-overlay` to create an overlay. 

Couldn't use the `::after` solution from `/articles` because it resulted in a weird gradient overlay on the shared header for a brief second when loading. 

Couldn't prepend `linear-gradient` to the inline background style because the site thought this was part of the image file name and returned a 404.


zarahzachz:
@seancdavis Yeah, I tried adding the linear-gradient to the end of the background. Also tried changing the property from `background-image` to `background`, no dice on it working. No worries though as the `bg-overlay` div did the trick and since it's a solution we're using elsewhere in the site it didn't take too much to implement.

I modified the background gradient on /articles to be a bit more pronounced. It required some magic numbers, but all screen sizes have been accounted for.

Ready to review.

zarahzachz:
@seancdavis I added the `bg-overlay` div to the jumbocards on the homepage.

Ready to review.

PR #257, opened 2018-07-24
Defect/de5678 related video images

zarahzachz:
### Problem
Recent/related video images aren't sizing appropriately on load. Also confirmed with Becky that these videos are not supposed to be carousel images on mobile.

### Solution
Change markup to traditional card grid layout.  This markup change seems to fix the image resize issue. Double-checked by slowing down the network speed and running a capture filmstrip thingy on Chrome - images come in at the correct size.

Test: `/videos/dude-this-is-a-really-nice-vidyo`

### DIFF
--- a/_includes/_video-card.html
+++ b/_includes/_video-card.html
@@ -6,7 +6,6 @@
 {% endif %}
 
 <div class="card">
-    <p>{{ video.date | date: '%B %d, %Y' }}</p>
   <a class="relative block" href="{{ video.url }}">
     {% if include.imgix_size == "1x" %}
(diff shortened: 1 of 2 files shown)


zarahzachz:
@seancdavis I ran through most of your feedback.
- I tested `sort: 'date'` and as I remembered, it does nothing. I think collections are sorted chronologically by default (oldest to newest), so this might just be a redundant filter here
- We actually need 9 items. I changed the limit to 9 and moved the slice filter to the end of the parameters. I think what was happening is that by slicing up to 9 items *then* excluding an item, we were always ending up with an 8-item array
- `exclude: page.title` does remove the current page from the array
- Changing limited.size to `== 0` instead of `< 1`

Ready to review.

PR #258, opened 2018-07-24
US14061 - Flagged Video Content

seancdavis:
Only show featured tags in the video sidebar.

Also fixes an issue with jekyll-placeholders that doesn't account for a collection not having a permalink.

---

Corresponds to crdschurch/crds-contentful-migrations#10.


vendor/gems/video-tags/spec/unit/generator_spec.rb
```diff
@@ -5,18 +5,15 @@
...
     @generator.generate(@site)
   end
 
   it 'should generate pages for video tags' do
     urls = %w(
-      /videos/tags/another-collection/index.html
-      /videos/tags/awesomeness/index.html
-      /videos/tags/collection-1/index.html
-      /videos/tags/culture/index.html
+      /videos/tags/collection-01/index.html
```
zarahzachz: Hey, I expected to see `featured-but-no-videos` listed here. Can you explain why there's only 1 value in this `%w` thing?

seancdavis: The featured item that has no videos doesn't get shown because it doesn't have any videos. The sidebar is populated by tags that are flagged as featureable _and_ that have been applied to at least one video.

seancdavis: For this odd syntax, here's a good summary: https://simpleror.wordpress.com/2009/03/15/q-q-w-w-x-r-s/

seancdavis:
@zarahzachz: Looks like the change I made to jekyll-placeholders was already accounted for in the abstracted gem. So all I did was remove it from the project.

PR #269, opened 2018-07-30
Map series date to 'start_date' and add sorting filters to the Recent…

zarahzachz:
### Problem
On the messages templates, Recent Series are sorting chronologically by "date". Because we aren't explicitly setting a date value in the config file, it's using today's date as the value resulting in a screwy array sort.

### Solution
Change the array sort to sort by `start_date`, and reverse it so it pulls the most recent values first. Update the config file to map the `date` field to `start_date` because Jekyll likes using `date` as a default value in a lot of places and this will prevent other bugs from cropping up.


_layouts/video.html
```diff
@@ -158,12 +158,12 @@ <h2 class="component-header flush-top">
...
         {% if page.series_slug %}
 
-        {% assign limited = site.series | sort: 'date' | reverse | slice: 0, 7 | exclude: page.series %}
+        {% assign limited = site.series | sort: "start_date" | reverse | exclude: page.series | slice: 0, 7 %}
         <section>
           <h2 class="component-header flush-top">Recent Series</h2>
           <div class="cards-2x">
             <div class="row">
-              {% for item in limited limit: 6  %}
+              {% for item in limited limit: 6 %}
```
lukemoderwell: do you need to `limit: 6` this if you've already sliced it above?

_config.yml
```diff
@@ -232,6 +232,7 @@ contentful:
           slug: slug
           image: image/url
           bg_image: background_image/url
+          date: starts_at
```
lukemoderwell: We're there other places we use this date property? I notice that below you switch to `start_date`...

_layouts/video.html
zarahzachz: I've been curious about slicing vs. limiting myself. We seem to be using a combination of the two throughout the site and I don't know the pros and cons between the two. In this instance I just fixed some white-space and assumed it was added here for a reason?

_config.yml
zarahzachz: This was added as a fallback/mapping thing. Basically, Jekyll likes using `date` as a default sort on arrays, so I mapped it to the same value as our `starts_at` value. The reason I kept the array sort as `starts_at` was I thought it was more readable to what was happening to that array.

_layouts/video.html
lukemoderwell: Slicing is generally good when you want items that don't start at the beginning (i.e. items 2 - 5). Limiting I believe is for when you want a group starting from the first item.

I would say pick one or the other. In this example, I think they both do the same thing.

PR #270, opened 2018-07-30
Defect/de5707 buy block layout

zarahzachz:
### Problem
Buy/share links on an album page are too close to the left side of their container. The [design](https://app.zeplin.io/project/5acf6d62553bdae0669efdd2/screen/5b2166d9532158cc60d3d2d8) calls for these links to be more centered.

### Solution
Add `.text-center` to center the content of this container and removed related flex CSS.


seancdavis:
@zarahzachz: This is really great. Removing unnecessary code to cleanup rather than building on top of previous solutions. Really good job here. 😄

PR #271, opened 2018-07-31
DE5595 - Small Background Image

seancdavis:
Implements [imgix-optimizer](https://github.com/ample/imgix-optimizer) as an external dependency. 

- 0.0.2 from this library implements looking at DPR to set proper background image size.
- 0.0.3 sets position automatically to fix a flashing issue.


_assets/javascripts/vendor/imgix-optimizer.js
```diff
@@ -0,0 +1,673 @@
...
+      classCallCheck(this, ImgixBgImage);
+
+      // Length of time to complete fade-in transition.
+      this.timeToFade = 500;
+      // Device pixel ratio assumes 1 if not set.
+      this.dpr = window['devicePixelRatio'] || 1;
+      // The primary element (i.e. the one with the background image).
+      this.el = $(el);
+      // Background image CSS property must be present.
+      if (this.el.css('background-image') == 'none') {
```
lukemoderwell: Does `this.el.css` also see inline styles? Lots of bg-imgs have traditionally been inlined.

_assets/javascripts/vendor/imgix-optimizer.js
```diff
@@ -0,0 +1,673 @@
+(function () {
```
lukemoderwell: should this file be minified like some of the other vendor js we pull in?

seancdavis: @lukemoderwell: Pushed another commit. We had compression enabled, but it wasn't working and it's not well-documented, so this fixes it. And because compression is enabled, the vendor files don't need to be because they will be compressed during build. And I like this better because a source map will also be provided in dev providing an ability to debug if necessary.

_assets/javascripts/vendor/imgix-optimizer.js
seancdavis: It sure do.

PR #291, opened 2018-08-16
Feature/us14577 topic detail markup

zarahzachz:
### Problem
Design for topic landing page outdated. It should match new designs found [here](https://app.zeplin.io/project/5acf6d62553bdae0669efdd2/screen/5b72fb36fc5a291a3fd88663)

### Solution
This story governs adding markup/styles - so there is no logic in place.


seancdavis:
Closing this since #295 contains the logic as well as these commits.

PR #293, opened 2018-08-17
DE5768 - Message breadcrumb

zarahzachz:
### Problem
The header for messages doesn't wrap correctly on mobile.

### Solution
Refactor header of videos/messages and add styles supporting mobile layout. I made the header elements stack on mobile and expand to an inline layout on tablet+ sized screens.

### Test
`/series/spark-2018/week-1-how-do-we-think-about-depression`

### DIFF
--- a/_assets/stylesheets/pages/_video.scss
+++ b/_assets/stylesheets/pages/_video.scss
@@ -15,14 +15,14 @@
   }
 
   header {
-    > div:first-of-type {
+    .sources {
       > a {
         display: block;
         font-size: $font-size-small;
       }
     }
 
-    > div:last-of-type {
+    > div:not(.sources) {
       margin-top: 12px;
     }
   }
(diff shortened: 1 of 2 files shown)


_assets/stylesheets/pages/_video.scss
```diff
@@ -13,4 +13,35 @@
...
+  @media screen and (min-width: $screen-xs) {
+    header {
+      display: flex;
+      justify-content: space-between;
+      margin-top: 2rem;
+
+      > div:first-of-type {
+        flex: 0 0 auto;
+        order: 2;
+        text-align: right;
```
galiotoweb: I think this is causing breadcrumbs to be right-aligned when there's no audio or program link:
screen shot 2018-08-17 at 11 11 20 am

zarahzachz: oh... cuz it becomes the `div:first-of-type` when the other div is gone. fixing it!

PR #308, opened 2018-08-29
Defect/de5823 signin signout redirection

jasoncallibrity:
Fix for /signin redirection!

### DIFF
--- a/_assets/javascripts/lib/set-redirect-url.js
+++ b/_assets/javascripts/lib/set-redirect-url.js
@@ -1,0 +1,3 @@
+window.onload = function() {
+    document.cookie = "redirectUrl=" + encodeURIComponent(window.location.href) + ";domain=.crossroads.net;Path=/";
+};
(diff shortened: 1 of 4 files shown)


_assets/javascripts/lib/set-redirect-url.js
```diff
@@ -0,0 +1,3 @@
+window.onload = function() {
+    document.cookie = "redirectUrl=" + encodeURIComponent(window.location.href) + ";domain=.crossroads.net;Path=/";
```
tcmacdonald: 1). Looks like your IDE is set to 4 spaces per tab. We prefer 2 spaces per tab. 
2). Do you actually need to set `Path=/` here?

jasoncallibrity: 1. Fixed.
2. Yes. If you don't set Path=/ it will write a new value for redirectUrl stacked in an array. Most likely wouldn't hurt anything but any length of navigation around the pages would produce a really ugly looking set of redirectUrl entries in the .crossroads.net cookie. Angular is the only piece of our stack that looks at this redirectUrl and path isn't used for processing in any way. Setting path is a way for document.cookie to overwrite it as the same cookie. Ultimately, when signin is updated this will need to be addressed if this is changed.

PR #311, opened 2018-08-30
DE5703 - Article metadata

zarahzachz:
### Problem
Meta descriptions in `<head>` should reflect the first line of text, regardless of whether it's lead or body copy.

### Solution
Update description metadata to use lead text if available, followed by an excerpt of the post content, and if neither is available - use the site description.

### DIFF
--- a/_includes/_head.html
+++ b/_includes/_head.html
@@ -28,23 +28,9 @@
   {% asset application.scss !integrity %}
 
   <title>{% if page.title and page.title != site.title %}{{ page.title | escape }} | {% endif %}{{ site.title | escape }}</title>
-  <meta name="description" content="
-  {% if page.lead %}
-    {{ page.lead | strip_html | truncate: 160 }}
-  {% elsif page.excerpt %}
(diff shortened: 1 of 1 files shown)


_includes/_head.html
```diff
@@ -28,9 +28,23 @@
...
+  {% endif %}">
   <meta property="og:title" content="{% if page.title %}{{ page.title | escape }}{% else %}{{ site.title | escape }}{% endif %}">
-  <meta property="og:description" content="{% if page.excerpt %}{{ page.excerpt | strip_html | strip_newlines | truncate: 160 }}{% else %}{{ site.description }}{% endif %}">
+  <meta property="og:description" content="
+  {% if page.lead %}
+    {{ page.lead | strip_html | truncate: 160 }}
+  {% elsif page.excerpt %}
+    {{ page.excerpt | strip_html | truncate: 160 }}
+  {% else %}
+    {{ site.description | strip html | truncate: 160 }}
```
seancdavis: ERRRRRR

PR #317, opened 2018-09-07
changed music landing page to use background image

bzmillerboy:
DE5871


tcmacdonald:
@bzmillerboy Looks good. Lets make sure we cherry-pick these commits back into `development` after you deploy the hotfix to production.

PR #359, opened 2018-10-04
Feature/us15215 download links

zarahzachz:
### Problem
Listen to audio link doesn't work as expected and should be removed. Download audio and video links should be available.

### Solution
I pulled out the listen to audio link and added download links for both audio and video. Since the original design called for stacking the listen to audio link on top of the program URL, I maintained that pattern. I replaced the listen to audio link with a download audio link and added a download video link beneath. If a program URL is provided, it will show up beneath the other two links (if applicable).

### Testing
`/series/empires/empires-week-1`


zarahzachz:
@tcmacdonald I am using the `download` attributes on the video and audio files. It looks like default behavior on browsers is to open a new tab for audio, video and images, where `txt` files trigger the force download window.

PR #370, opened 2018-10-09
DE6000 - Sign out button

zarahzachz:
### Problem
Account btn on mobile nav to nav to profile or sign in depending on login status.

### Solution
Wrap account btn in conditional logic to show/hide certain elements if validated.


seancdavis:
@zarahzachz: You have a conflict here. Also, I'm unable to get this working for me. It's just redirecting to int.crossroads.net.

Also, I just noticed there's no way to sign out on mobile.

zarahzachz:
@seancdavis I resolved the merge conflicts. And as far as checking the redirects, I _think_ Taylor said the auth stuff only worked when the site was served from a CR domain name. I might be mistaken here, but I figured these redirects weren't working for that reason.

PR #373, opened 2018-10-10
US14987 - Menu Squencher

seancdavis:
Adds automatic menu squenching for all your squenching needs.


zarahzachz:
@seancdavis looks great! Couple of thoughts:
1. I thought Becky said she wanted the nav to collapse until just Articles remained?
2. The 'All Media' link sits about 2 px higher than the rest of the links inline with it.

seancdavis:
Changes made.

I also bumped the padding to `28px` in preparation for DE6008 which will make the side padding `.75rem` for menu items on larger screens.

PR #381, opened 2018-10-11
Feature/us15267 nav active state

zarahzachz:
### Problem
Nav links don't have an active state when navigating through site.

### Solution
Add the active state to links using liquid. I tried to put this logic inside a `{% capture %}` block var to clean it up some, but this messed with the data assigned to the url `{% assign %}`s


_includes/_desktop-header.html
```diff
@@ -12,15 +13,29 @@
...
     <ul class="list-inline">
       {% for topic in data.topics %}
-      <li><a href="{{ topic.url }}">{{ topic.title }}</a></li>
+      {% assign topic_url = topic.url | split: '/' %}
+      <li>
+        {% if topic_url == page_url %}
+        <a href="{{ topic.url }}" class="active">{{ topic.title }}</a>
+        {% else %}
+        <a href="{{ topic.url }}">{{ topic.title }}</a>
+        {% endif %}
```
seancdavis: Feels like a lot of extra code, when you could accomplish this like so:

```liquid
<a href="{{ topic.url }}" class="{% if topic_url == page_url %}active{% end %}">{{ topic.title }}</a>
```

_includes/_desktop-header.html
```diff
@@ -12,15 +13,29 @@
   <div class="media-header-section">
     <ul class="list-inline">
       {% for topic in data.topics %}
-      <li><a href="{{ topic.url }}">{{ topic.title }}</a></li>
+      {% assign topic_url = topic.url | split: '/' %}
```
seancdavis: What is breaking these up into array accomplishing? Don't you end up with the same values? Does liquid remove nil items from the array?

zarahzachz: I was trying to come up with a way to compare the link url with the page url, regardless of whether either link ends with a `/`. I figured by converting both values to arrays without a slash it was the easiest way to compare the real value without worrying about errant slashes.

_includes/_desktop-header.html
zarahzachz: Oh, that's good. Updating!

PR #412, opened 2018-10-24
Defect/de6053 duration display

zarahzachz:
### Problem
Messages not receiving an icon on the homepage.

### Solution
Add messages to the media label partial.

### Testing
Visit the home page and see if recent media cards (specifically messages) have an icon on them.

**Bonus fix:** I also fixed the duration logic. Looks like the content model changed since this partial was written, media types either have a `audio_duration` or `video_duration` field and not a `duration` field. I assume this was to account for media types with both? I didn't see this on Contentful, but could've missed it. 

Also added the "min watch/listen" text to the div to match the style of articles. I don't know if this was how it originally looked but some of the cards looked strange with a single number just kinda hanging out there.

### DIFF
--- a/_includes/_media-label.html
+++ b/_includes/_media-label.html
@@ -13,8 +13,10 @@
       <span class="font-size-smallest">{{ read_time }} min read</span>
     {% endif %}
   {% elsif content_type == "video" or content_type == "message" %}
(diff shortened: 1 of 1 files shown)


zarahzachz:
@seancdavis I updated the logic. Episodes are the only media type that uses integers on the duration field, so outside of articles, that's the only type to have the "min watch" label. I've removed the label from the other media types.

Ready to review.

PR #457, opened 2018-12-12
US15878 Submodules

tcmacdonald:
Implement non-tracking orphan branch `_includes` as submodule so the same files can be used by crdschurch/crds-net#394


seancdavis:
Closing this (for now). We're piloting submodules in crds-net with just the `_pages` directory. We will come back to this approach eventually, but it may be slightly different at that time, based on the outcome of the pilot.

PR #464, opened 2018-12-17
Feature/us15523 overlay card documentation

zarahzachz:
## Problem
Overlay cards are locally styled.

## Solution
Remove locally defined overlay card styles from project and add to DDK.

### Corresponding Branch
crdschurch/crds-styles#352
crdschurch/crds-styleguide#386

## Testing
A simple test would be to see if the homepage looks goofy.


zarahzachz:
Preview links won't work for media for this PR. Must be built locally to see styles are being pulled in correctly.

PR #536, opened 2019-03-01
Feature/us16271 media jumbotron content block

phritolay:
## Problem
heading and description can not be updated in Contentful

## Solution
update heading and description to pull from Contentful Content Block

### Corresponding Branch
https://github.com/crdschurch/crds-contentful-migrations/pull/96


_plugins/filters/get_doc_filter.rb
```diff
@@ -5,7 +5,7 @@ def get_doc(obj)
       begin
         return nil if obj.nil?
         return get_docs(obj) if obj.is_a?(Array)
-        doc = collection_from_obj(obj).detect { |doc| doc.data['id'] == obj['id'] }
+        doc = (collection_from_obj(obj) || []).detect { |doc| doc.data['id'] == obj['id'] }
```
seancdavis: Nice fix here! Thank you.

index.html
```diff
@@ -32,8 +34,8 @@
   <div class="header-content container">
     <div class="row">
       <div class="col-md-8 col-lg-9 text-left">
-        <h1 class="title-xlg font-family-condensed-extra text-uppercase push-quarter-bottom">Another Perspective</h1>
-        <p class="font-family-base font-size-base">Better answers to the questions we're all asking.</p>
+        <h1 class="title-xlg font-family-condensed-extra text-uppercase push-quarter-bottom">{{ jumbotron.title }}</h1>
```
seancdavis: Interesting way to do it using the title and description together.

Question: It looks like you're referencing `description` directly and not parsing for markdown. I'm guessing that means it won't render markdown? You can either use the `markdownify` filter or use the `content` attribute, which _should_ render the markdown automatically.

However, that causes problems. You are placing this text within a `<p>` tag and rendering markdown will also generate paragraph tags. Could you wrap within a `<div>` instead, without losing the appropriate styles?

phritolay:
@seancdavis I can't replicate the hubspot caption issue locally or on the preview build.

screenshot 2019-03-01 16 31 18

index.html
phritolay: Thanks Sean, pushing up a fix now.

phritolay:
I saw the hubspot caption issue on the latest preview, but once I refreshed it was fine.

PR #547, opened 2019-03-12
us16605 add feature content section to homepage

phritolay:
## Problem
Featured media wasn't displaying when content was in Contentful

## Solution
update featured media to pull in 'site' field from Contentful
update feature media to render content from Contentful when present

### Corresponding Branch
https://github.com/crdschurch/crds-net/pull/748
https://github.com/crdschurch/crds-contentful-migrations/pull/102


index.html
```diff
@@ -19,8 +19,9 @@
     sort: starts_at desc
 ---
 
-{% assign feature_obj = site.featured_media | where: 'page_path', page.url | first %}
-{% assign managed_features = feature_obj.docs %}
+{% assign feature_obj = site.featured_media | where: 'site', 'media' | where: 'page_path', page.url | first %}
```
seancdavis: This is fine in this case, but we can fix it further up the chain to prevent us from running into this issue again.

Notice how we're [filtering the content block query](https://github.com/crdschurch/crds-media/blob/development/_config.yml#L120-L121), such that only those scoped to the media project will be brought it. This decreases build time and removes the need to add `where: 'site', 'media'` to every place we use features on this site.

index.html
```diff
@@ -19,8 +19,9 @@
...
+{% assign managed_features = feature_obj.entries %}
```
seancdavis: Recall talking through what `get_docs` does.

In this case, `feature_obj.entries` is just data object (a Ruby hash), which is why you can call things like `.title` and other fields. _But_, there is no `.url` method because Jekyll is responsible for those and Jekyll does not know about the association, only the raw data.

That's why we pipe to the `get_docs` filter in instances like this. It finds the appropriate Jekyll object so you can call necessary methods, like `.url`.

We can walk through an example, but if you test your code on the front-end, you'll notice clicking on the features only reloads the current page.

seancdavis: Actually, in pairing with Warren, I thought about this and `get_docs` is not the right approach here.

Instead, check out the features generator (in `_plugins/generators`). It is looping through the featured_media objects and adding a `.docs` array that stuffs in all the appropriate Jekyll objects.

Long story short—change `.entries` to `.docs` and it should work. But I also want to make sure you understand what you're doing and why you're doing it. So I'm happy to walk through it with you.

index.html
phritolay: 👍 That's awesome. I'll update it.

index.html
phritolay: I ended up changing it to `.entries` because when it was set to `.docs` only the 4 most recent posts were displaying.  I just tested it again and I'm getting the same result. 🤔 I might be missing something, so I think it would be good to walk through it.

PR #562, opened 2019-03-20
US14397 Message Time Markers

tallented:
## Problem
There isn't a way to link directly to a time marker in a video

## Solution
* Added JSON Editor extension to Contentful Message and Video content Model
* One or more time markers can be added
* This creates a Content tab on the Message and Video layout in Jekyll
* Clicking a time marker also adds the min/sec to the browser address bar URL
* If the page loads and the message/video URL has a min/sec time marker on it, then the video will autoplay from that position 
* Clicking "Listen to Audio" will stop the playing video
* Clicking "Watch Video" will stop the playing audio

### Corresponding Branch
*Add link to crdschurch/repo-name-here#issue-number (if applicable). This helps the code reviewer know that corresponding work exists and where to find it.*

## Testing
Example links using the Contentful environment `featureUS14397timemakerlinks`:
* /videos/insights-why-would-a-church-do-the-super-bowl-of-preaching?min=4&sec=30
* /series/spark-2018/week-5-ideas-to-ignite-your-life?min=3&sec=43


_assets/javascripts/components/audio_video_toggler.js
```diff
@@ -10,21 +10,21 @@ class AudioVideoTrigger {
     this.playerContainer = $('[data-audio-video-control]').first();
     this.audioTrigger = $('[data-audio-trigger]').first();
     this.audioPlayer = $('[data-audio-player]').first();
-    if (this.audioPlayer.length > 0) this.audioHTML = this.audioPlayer[0].outerHTML;
```
tallented: The Youtube iframe player API needs to stay attached to the DOM, so the original mechanism of removing the player and replacing it won't work.

_includes/_tabs.html
```diff
@@ -2,15 +2,28 @@
   <ul class="nav nav-tabs" role="tablist">
   {% for tab in include.tabs %}
     {% if page[tab] != blank %}
-    <li role="presentation"><a href="#{{tab}}" aria-controls="{{tab}}" role="tab" data-toggle="tab">{{ tab | capitalize | replace: "_", " " }}</a></li>
+    <li role="presentation"><a href="#{{tab}}" aria-controls="{{tab}}" role="tab" data-toggle="tab">
```
tallented: Optionally pass in a name for the tab from the calling page, otherwise capitalize the tab ID.  This is necessary because the time marker tab needs to be named Contents but using contents as the Content Model field name wasn't working so I used timestamps instead.  It appears that Contentful uses `contents` field for a different purpose

_includes/_tabs.html
```diff
@@ -2,15 +2,28 @@
...
     {% endif %}
   {% endfor %}
   </ul>
 
-  <div class="tab-content soft-top">
+  <div class="tab-content">
   {% for tab in include.tabs %}
     <div role="tabpanel" class="tab-pane soft-bottom" id="{{tab}}">
-      {{ page[tab] | markdownify }}
+      {% if include.templates[forloop.index0] and include.templates[forloop.index0] != 'none' %}
```
tallented: The tabs include currently assumes tab content is just Markdown on the page front-matter.  Since the video time markers is a completely different approach I've externalized the HTML into a template file.  The calling page can use the liquid template split workaround to create an array of file names to use for the tab layout instead of markdown

_includes/_yt-embed.html
```diff
@@ -1,9 +1,48 @@
...
+          height: '315',
+          width: '560',
+          videoId: '{{ video_id }}',
+          events: {
+            'onReady': onPlayerReady
+          }
+        });
+      };
+    }, 0);
+
```
tallented: Checks to see if there is a ?min=X&sec=Y on the URL and if so, seeks to that position in the video and autoplays

PR #570, opened 2019-03-21
Feature/us16760 home featured collections

hungrymedia:
## Problem
Featured collections are not visible on home page

## Solution
Display featured collections in sidebar of home page.

### Corresponding Branch
- Uses collection flag introduced in crdschurch/crds-contentful-migrations/tree/feature/US16760-home-featured-collections
- Depends on crdschurch/crds-media/pull/568

## Testing
As a crds-media editor
When I choose to feature a collection on the homepage
Then I should see that (or those) collections on the homepage


seancdavis:
Don't merge this before #568 so that we aren't clicking into pages that don't exist.

PR #580, opened 2019-03-26
DE6595 / DE6596 - Featured collection display in media landing page

hungrymedia:
## Problem
On the media landing page, incorrect image was displaying for featured collections. Additionally, the story count was missing under the collection information.

## Solution
- A new field was added to collections to allow for a secondary image
- Story count was added

### Corresponding Branch
https://github.com/crdschurch/crds-contentful-migrations/pull/115


hungrymedia:
> 1. Need some space between the description and the story count. It looks like it's `6px` in the design.

Per our discussion, spacing matches existing style on the match, so will be left, as is.

seancdavis:
@JenEllis44 @beckysingson I merged crdschurch/crds-contentful-migrations#115 (it's a migration, not front end change).

Warren did not change the spacing I suggested here to stay consistent with what we have on the rest of the page.

We will now have a field called `secondary_image` which will populate the image on the home page. In its absence, we fall back to the image (image being that which is used as the background on the detail page).

PR #606, opened 2019-04-04
DE6662 - Hubspot form submission message centering.

hungrymedia:
## Problem
Message returned after Hubspot form submission was not centered with surrounding content.

## Solution
Centered `<p>` tags within `.submitted-message` area.


hungrymedia:
@JenEllis44 This is happening, because Hubspot is returning a bunch of blank `<p>` tags with nothing but `&nbsp;`. Since they actually contain content, albeit a space, they are given a line-height. I don't see a way to handle this with purely CSS, but perhaps I can target them via Javascript and then collapse them. @seancdavis Can you think of another way to hide these?

image

hungrymedia:
@JenEllis44  The hubspot subscription form has been updated to use an inline response message, per [Hubspot's form option docs](https://developers.hubspot.com/docs/methods/forms/advanced_form_options). I also broke the form out into a partial, since it is also used on the media landing page (which was seeing the same issue):

https://5caba526eba254000acb2378--crds-mediaint.netlify.com/collections/community-stories

hungrymedia:
@seancdavis I'll see what Hubspot allows for overriding errors. I should be able to match the error style, visually, but markup is coming from Hubspot, so it won't match the DDK exactly.

PR #660, opened 2019-05-29
customize quality labels

lukemoderwell:
## Problem
Bitmovin's generic quality labels suck

## Solution
add human readable dimensions and bitrates


lukemoderwell:
@janssejl @j-brammer Brian had originally asked to hide certain streams but that's just too complex for the value right now.

PR #668, opened 2019-06-04
[CYP] US17354 - Bitmovin autoplay/subtitles tests

s-henney:
## Test coverage added
- Tests that a Bitmovin message is autoplayed with/out sound based on query params.
- Tests that subtitles are displayed when autoplaying without sound.
- Uses Amplitude events to check if a Bitmovin player is playing, so tests those are firing too.


cypress/integration/messagePage/bitmovin_message_autoplay_subtitle_mute_spec.js
```diff
@@ -0,0 +1,155 @@
+import { MessageQueryManager } from '../../Contentful/QueryManagers/MessageQueryManager';
+import { AmplitudeEventChecker } from './helpers/AmplitudeEventChecker';
+import { BitmovinPlayer } from './helpers/BitmovinPlayer';
+
+const soundOn = 'sound=11';
+const soundOff = 'sound=1';
+describe('Tests Bitmovin message with transcription autoplays with/out subtitles based on query params', function () {
```
j-brammer: carriage return between 6 and 7?

cypress/integration/messagePage/bitmovin_message_autoplay_subtitle_mute_spec.js
```diff
@@ -0,0 +1,155 @@
...
+  let messageWithSubtitles;
+  before(function () {
```
j-brammer: carraige return here?

I think VSCode has a code autoformat command you could use to fix many of those issues. On Mac its control+shift+f.

PR #669, opened 2019-06-05
uncommenting autoplay

janssejl:


_includes/_video-player-card.html
```diff
@@ -1,6 +1,6 @@
-<!-- {% assign video = item %}
+{% assign video = item %}
     {% if video.bitmovin_url and include.autoplay == true %}
```
j-brammer: if and endif are out of alignment

PR #675, opened 2019-06-07
align bitmovin ui with brand blue

lukemoderwell:


bzmillerboy:
Are lines 57 and 55 the same?

lukemoderwell:
@bzmillerboy good catch. Looks like they are.

PR #688, opened 2019-06-13
updating jekyll-crds to fix music autoplay params

janssejl:
checking for related_videos for media types and adding autoplay params if bitmovin url exists


janssejl:
and adding autoplay to the collections page for overlay cards

PR #713, opened 2019-06-26
US17591 - chaning bitmovin subtitles to be added externally from contentful

janssejl:
Has other PRs on CRDS-NET and CRDS-VIDEO-SERVICE


_includes/_bitmovin-player.html
```diff
@@ -37,7 +37,8 @@ <h2 class="overlay-card-title">{{ item.title | truncate: 55 }}</h2>
         url: `{{ video.bitmovin_url }}`,
         id: videoManagerID,
         image: `{{ video.image.url | imgix: site.imgix }}`,
-        isCard: `{{ include.isCard }}`
+        isCard: `{{ include.isCard }}`,
+        subtitles_url: '{{ video.transcription.url }}'
```
j-brammer: Won't this cause issues if there isn't a transcription url? Do we need to do the following?

```suggestion
        {% if video.transcription.url %}
        subtitles_url: '{{ video.transcription.url }}'
        {% end if %}
```

PR #737, opened 2019-07-10
Feature/us17681 remove songs album

J-Byron:
- Removes song/album styles and markdown 
- Redirects music refs to https://www.crdsmusic.com/
- Adds query to filter songs that don't include videos

## Testing
- https://deploy-preview-737--crds-mediaint.netlify.com/

PR #738, opened 2019-07-10
US17682-music-redirects

J-Byron:
- Changes /music refs to crdsmusic

## Testing
- https://app.netlify.com/sites/crds-mediaint/deploys/5d25f2a7ca56750008067d6e


_includes/_shared_global_header.html
```diff
@@ -40,7 +40,7 @@
 <ul><li class="dropdown-header">Media Categories</li>
 <li><a href="//int.crossroads.net/media" data-automation-id="sh-allmedia">All Media</a></li>
 <li><a href="//int.crossroads.net/media/series" data-automation-id="sh-series">Series</a></li>
-<li><a href="//int.crossroads.net/media/music" data-automation-id="sh-music">Music</a></li>
+<li><a href="https://www.int.crdsmusic.com/" data-automation-id="sh-music">Music</a></li>
```
seancdavis: This file is getting removed via #737.

_redirects
```diff
@@ -1,4 +1,7 @@
-/albums /music 302
+/albums/* https://www.int.crdsmusic.com/albums/:splat 302
+/songs/* https://www.int.crdsmusic.com/songs/:splat 302
+/music https://www.int.crdsmusic.com/ 302
```
seancdavis: Two things here:

1. `www` is prod, `int` is for int, `demo` for demo.
2. Because of that (☝️)  we can't write this directly into the project. Instead I'm thinking we're going to have to write the redirects into the [crds-net file](https://github.com/crdschurch/crds-net/blob/development/redirects.csv), which has environment support.

index.html
```diff
@@ -162,7 +162,7 @@ <h2 class="collection-header push-bottom text-gray-light">by media type</h2>
         <div class="card-deck">
           <ul class="list-unstyled font-size-base flush-bottom">
             <li><a href="/series">Series</a></li>
-            <li><a href="/music">Music</a></li>
+            <li><a href="https://www.int.crdsmusic.com/">Music</a></li>
```
seancdavis: This must programmatically account for the correct subdomain.

_data/navigation.yml
```diff
@@ -62,7 +62,7 @@ megamenu:
       url: /series
 
     - title: Music
-      url: /music
+      url: https://www.int.crdsmusic.com/
```
seancdavis: This must programmatically look for the right subdomain. This could be tricky. We'll probably want to talk through this.

PR #741, opened 2019-07-12
DE6919-trending-topics-bugs

J-Byron:
## Problem
- "More related topics" cards were not loading card data correctly

## Solution
- Properly sets up media-card to consume card data


_layouts/tag.html
```diff
@@ -40,7 +40,7 @@ <h2 class="collection-header text-gray-light text-lowercase push-bottom">
     <div class="cards-4x" data-page="tagged_media">
       <div class="row" data-page-number="{{ page.tagged_media.page }}">
         {% for card in page.tagged_media.docs %}
```
lukemoderwell: @J-Byron curious if you thought about changing this to `item` instead of passing it in as card. Is something else relying on the `card` var?

J-Byron: @lukemoderwell I am quite new to liquid and wasn't thinking of this solution at the time. Looking back this makes much more sense.

PR #748, opened 2019-07-17
[DH-IP] Off-cycle Prod Deploy

seancdavis:


_config.yml
```diff
@@ -99,6 +99,7 @@ contentful:
     map:
       date: published_at
   episode:
+    query: content_type=episode&fields.podcast.sys.contentType.sys.id=podcast&fields.podcast.fields.exclude_from_crossroads[ne]=true
```
j-brammer: build is failing due to the contentful migrations not being run in prod yet I believe. Approval with the caveat that the build must be successful before merge.

PR #796, opened 2019-10-21
US18234-exclude-IGR-categories

J-Byron:
### Problem 
- Categories that belong to IGR collection should be excluded from Media
- IGR collection object should also be removed


_config.yml
```diff
@@ -132,9 +132,11 @@ contentful:
     map:
       description: content
   collection:
-    query: fields.slug[ne]=briantome-com
+    query: fields.slug[nin]=briantome-com,isgodreal-com
```
tcmacdonald: This feels brittle to me. What happens when they add new collections?

_config.yml
```diff
@@ -132,9 +132,11 @@ contentful:
     map:
       description: content
   collection:
-    query: fields.slug[ne]=briantome-com
+    query: fields.slug[in]=crdsmusic-com,
   podcast:
     query: fields.exclude_from_crossroads[ne]=true
+  category:
+    query: "fields.collections.sys.id[nin]=3F1707Eeiu50rMaxCf6wNN,3X7oFhglWTJGXy3TD5u5br,6XwgUaVRdsKzXrncprwJMt,3E8JlwCIG49c00bU7MXEuL"
```
seancdavis: Same comments apply here as on the crds-net PR.

PR #811, opened 2019-12-19
Feature/us17184 lead text meta

silasbaur:
## Problem
For some reason, the description and og:description content are not using the lead text in the main body of the page but instead, skipping that and going to the second paragraph. It should include the lead text.

## Solution
Add logic that says if page content type = Article and page lead text != nil then use lead text as meta description
[Preview](https://deploy-preview-811--crds-mediaint.netlify.com/articles/the-people-executives-promote)


silasbaur:
@tcmacdonald Yes It is

tcmacdonald:
@silasbaur Please open a new branch sourced from `development` and cherry-pick 
781b86d and d12814d into that new branch and submit a PR to `development`.

PR #813, opened 2019-12-23
US17184

silasbaur:


tcmacdonald:
Thanks @silasbaur. I think a better approach would be to apply this business logic to the custom Jekyll filter that's already in place across a number of these sites. 

https://github.com/crdschurch/jekyll-crds/blob/master/lib/jekyll-crds/filters/meta_filters.rb#L20-L32

Closing this PR for now but let me know if you have any further thoughts.

PR #835, opened 2020-01-31
Adds updates for distribution channels

J-Byron:
## Problem
- Net needs to be updated to compensate for the incoming distribution channel changes

## Solution

- Removes config queries related to canonical_host
- Adds config for distribution channels site
- Bumps jekyll-contentful to <s>2.1.0</s> 3.0.0


_config.yml
```diff
@@ -87,18 +87,18 @@ collections:
     output: true
 
 contentful:
+  config:
+    sites: "www.crossroads.net"
```
tcmacdonald: Can you review the usage docs on this feature of jekyll-contentful? This does not appear to be an accurate use of this configuration value.

tcmacdonald: https://github.com/crdschurch/jekyll-contentful#content-distribution

J-Byron:
> jekyll-contentful version did not change in this PR.

Looks like Jason updated jekyll-contenful in development https://github.com/crdschurch/crds-media/pull/831

J-Byron:
Looks like Jason updated jekyll-contenful in development #831

PR #865, opened 2020-03-26
US19168 Ctfassets

silasbaur:
## Solution 
We start out with 4354 references to ctfassets and decided to update the build script with sed to change assets and images after the site is built. After running the sed command we are left with 3686 references to ctfassets, that is 571 downloads and 3110 videos. We are not changing assets because they are mostly PDFs and it changes the content type to an image when being hosted from Imgix so It does not get treated as an PDF(when clicked on, it won’t show print or download option like you would normally see from a PDF). We are also not changing videos because some of these videos are entire sermons so the file size is going to be very big and imgix will not be able to handle that.


bin/build-command.sh
```diff
@@ -6,6 +6,7 @@
   bundle exec jekyll contentful --sites www.crossroads.net &&
   bundle exec rspec ./vendor ./spec &&
   bundle exec jekyll build &&
+  find . -type f -name "*html" -print0 | xargs -0 sed -i '' -E 's/\/\/(assets|images).ctfassets.net\/y3a9myzsdjan/\/\/crds-media.imgix.net/g' &&
```
tcmacdonald: @silasbaur Please update this to ensure you're only replacing URLs that match images.ctfassets.net.

PR #879, opened 2020-04-22
update some error handling on crds-media test and skip some test th…

lynchl607:
update some error handling on crds-media test and skip some test with functionality change


lynchl607:
out of date with repo

PR #945, opened 2020-06-23
Production Deploy

tcmacdonald:
## Problem
*Document current behaviour and why it's undesirable here.*

## Solution
*Document how the behaviour was fixed here.*

### Corresponding Branch
*Add link to crdschurch/repo-name-here#issue-number (if applicable). This helps the code reviewer know that corresponding work exists and where to find it.*

## Testing
*Include information on how to test your work here (if applicable).*


_assets/stylesheets/pages/_author.scss
```diff
@@ -27,30 +27,30 @@
...
+    height: 30px;
+    justify-content: center;
+    margin-right: 10px;
+    width: 30px;
+
+    &:focus,
+    &:hover {
+      background-color: $cr-orange-dark;
+    }
+  }
```
silasbaur, line 45: extra spaces in this file

bin/build-command.sh
```diff
@@ -2,7 +2,6 @@
 # Ships logs to logz.io
 
 {
-  bundle exec jekyll crds &&
   bundle exec jekyll contentful --sites www.crossroads.net &&
```
j-brammer, line 5: Maybe this was accounted for, but I see that crds-net still has this build command. What is different about crds-media?

j-brammer: Looks like the issue is actually in crds-net, this command needs removed from there

PR #1031, opened 2020-12-21
Article background image refactor and alt text

jakecatron:


_layouts/article.html
```diff
@@ -6,15 +6,17 @@
 {% assign author = page.author | get_doc %}
 {% assign topic = page.category | get_doc %}
 {% assign tags = page.tags | get_doc %}
+{% assign alt = page.assets | alt %}
```
tcmacdonald, line 9: Is this necessary? I suspect this isn't actually returning anything.

_assets/stylesheets/pages/_articles.scss
```diff
@@ -20,6 +19,16 @@
         }
       }
     }
+    img {
+      height: 790px;
+      width: 100vw;
+    }
```
tcmacdonald, line 25: What if I upload an image to the CMS that is less than 790px tall?

_layouts/article.html
jakecatron: Yes, you are correct. Just removed that!

beckysingson:
@jakecatron @tcmacdonald Leaving comments here since this work doesn't have a corresponding Rally story at the moment:

The new jumbotron image doesn't respond appropriately on mobile. 

Here's how the deploy preview appears on Safari on iOS:
IMG_4559

Compared to how the article jumbotrons appear currently on INT:
IMG_4558

kellgauthier (approved):

PR #1035, opened 2020-12-22
Update gtag Implementation

tcmacdonald:


_assets/javascripts/components/analytics.js
```diff
@@ -0,0 +1,178 @@
...
+function gtag(){
+  a.gtag(arguments);
+}
+
+/**
+ * 
+ */
+function ga(originalCommand, event, eventCategory, eventAction) {
+  a.ga(originalCommand, event, eventCategory, eventAction);
+}
```
mjvezzani, line 178: Hold up. So we ended up having to wrap and make global both the gtag _and_ ga functions?

_assets/javascripts/components/analytics.js
```diff
@@ -0,0 +1,178 @@
...
+  this.gtag('config', 'UA-2977073-1', this.gtagConfig());
+  this.sendPageView()
+}
+
+/**
+ * 
+ */
+Analytics.prototype.gtag = function() {
+  dataLayer.push(arguments);
+}
```
mjvezzani, line 24: Help me understand what is going on here. Does this have something to do with providing GTM / GA dataLayer information?

Also, is it not necessary to provide that `arguments` variable as part of the method signature? (i.e. `Analytics.prototype.gtag = function(arguments)`)

mjvezzani (approved):
Looks good. I just had some confusion around the `gtag` prototype method definition.

_assets/javascripts/components/analytics.js
tcmacdonald: The `gtag()` was already defined as a global method [here](https://github.com/crdschurch/crds-media/blob/b74a365b20479635f185000acb883f6114e57866/_includes/_gtag.html#L5) so this PR just moves into the same file as the rest of our concerns.

_assets/javascripts/components/analytics.js
tcmacdonald: 1. All gtag.js does is use the `dataLayer` array as a means to push data into Google products ([more here](https://developers.google.com/tag-manager/devguide)). This is a drastically simpler product-agnostic solution that GTM has been using for years now. And if you scrutinize [your original implementation](https://github.com/crdschurch/crds-media/blob/b74a365b20479635f185000acb883f6114e57866/_includes/_gtag.html#L4-L7), you'll see how this works– nothing magical here. 

2. The `prototype` stuff is a result of ES5's protypal inheritance pattern ([more here](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Inheritance#Prototypal_inheritance)). Prior to ES6, the widely supported strategy for creating a OO class in JS follows this approach... 

```
function Cat() {
  // this is the constructor
}

Cat.prototype.sayHello = function() {
  // this is an instance method
}

new Cat().sayHello() // meow
```

3. I don't think you want to specify an actual argument called `arguments` because `arguments` is a reserved keyword and you may be messing up the native behavior by doing so ([moar here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments)).

PR #1045, opened 2020-12-23
Prod Deploy 12-23-20

silasbaur:


_includes/_head.html
```diff
@@ -13,6 +13,29 @@
     {% endif %}
   {% endif %}
 
+  {% if page.content_type == "article" %}
+  <title>{{ page.title }}</title>
```
tcmacdonald, line 17: Hold up. This will remove the title tag from all pages other than articles right?

_assets/stylesheets/pages/_articles.scss
```diff
@@ -20,6 +19,26 @@
         }
       }
     }
+
+    img {
```
tcmacdonald, line 23: This seems like a very broad declaration. What if I include an `img` within the content of an article?

_includes/_head.html
silasbaur: That is correct

tcmacdonald (changes requested):
Sorry I missed these two items in the previous PR. We need to inspect them and determine whether these can go to production.

_includes/_head.html
```diff
@@ -13,6 +13,29 @@
...
+    <script type="application/ld+json">
+    {
+      "@context": "https://schema.org",
+      "@type": "Article",
+      "headline": "{{ page.title }} ",
+      "image": [
+        "{{ page.image.url }}",
```
dholloran-ample, line 24: Schema validation fails here [see results](https://search.google.com/test/rich-results?utm_campaign=sdtt&utm_medium=url&id=VoYh23xyg8PBG-W8-g5Rig)

@silasbaur - I can look into this just wanted to make sure we didn't move forward

dholloran-ample (changes requested):
@silasbaur - Left one comment I am going to look into

_includes/_head.html
tcmacdonald: Good catch, Dan. I'm guessing it expects a string not an array. Thanks!

tcmacdonald: Confirmed... docs say this should be a URL.

https://schema.org/image

dholloran-ample (approved):

tcmacdonald (approved):

PR #1056, opened 2021-01-18
updates links

kellgauthier:
## Task
There are a few links which are pointing to prod instead of int - breaking automated tests.
([Rally](https://rally1.rallydev.com/#/38108142417d/custom/24109743995?detail=%2Fdefect%2F485860365812)) 

Links that need to be corrected are: 

- Download the app
- Check out our groups
- Try us for 30 days
- About Crossroads

In Searching for answers section:

- Spiritual outfitter 
- Making a difference 

_**Guide people is not clickable and cursor is not turning to hand when hovering around it
**Build community is not clickable and cursor is not turning to hand when hovering around it
**Questions is not clickable and cursor is not turning to hand when hovering around it_

****This is content that AG still owes us and will be linked at a later date.** 

## Testing
Will not work on deploy preview.


dholloran-ample (changes requested):
@kellgauthier - `{{ jekyll_env }}` does not seem to be outputting anything. I think it is `{{ site.jekyll_env }}` which seems to output `int`, `development`, etc. I'm not sure if this will work for what you're trying to do.

We also can not do just `/slug` since that ends up being `/media/slug` once it's merged in.

Sorry I can not provide more direction. I would think this has to have been solved before (quick search of media did not turn anything obvious up for me net may have soemthing).

PR #1057, opened 2021-01-19
DE8136/fixes styles on bottom subscribe section

kellgauthier:
## Problem
The design of the "Now Keep It Coming" subscribe pre-footer does not match [design](https://projects.invisionapp.com/d/main#/projects/prototypes/20464001). 
[Paymo Task](https://rally1.rallydev.com/#/66096747656ud/custom/24109743995?detail=%2Fdefect%2F485852334380)

## Testing
*Cannot recreate sidebar subscribe defect - seems resolved*


_assets/stylesheets/pages/_media-homepage.scss
```diff
@@ -176,3 +176,9 @@
     }
   }
 }
+
+.hs-richtext, .hs-main-font-element {
```
dholloran-ample, line 180: Are you trying to hide the "Get our top picks delivered to your inbox every week." block or something else? 

I'm concerned that the selectors are too generic and may hide items we do not intend to. Please add either a wrapper class to the form(s) or some other way to avoid collisions.

dholloran-ample (changes requested):
@kellgauthier - I left one piece of feedback below

dholloran-ample (approved):

PR #1058, opened 2021-01-21
DE8138 Media Feedback

silasbaur:
## Problem
- [ ] Why do we have "More about Brian Tome but not for the other authors? Is this intentional? Should this be more obvious? Should the other authors have this and just link to their author pages? And More about Brian does not look like link, but it's clickable 
- [ ] Get rid of "View All" We are just linking to the page they are on
- [ ] Under "Recent" in the design we list the author, but not on Int
- [ ] More space above the podcast/Music line in the tan box (like in the design) It's super close right now. 
- [ ] Inconsistent Spacing 
- [ ] On mobile What is the grey shadow cutting through the word "work" on the snail trial?
- [ ] On mobile make sure things line up


dholloran-ample (changes requested):
@silasbaur - 2 things below

1. "Inconsistent spacing" the spacing still seems off. I'm seeing `~50px` top and `~70px` bottom. There may be an extra margin or something.
2. Issue 9 about adding a link on the Podcast title to the podcasts page. Is this something that should be taken care of? The ticket is kind of hard to follow.

dholloran-ample (approved):

PR #1068, opened 2021-02-03
Prod Deploy 2-3-2021

silasbaur:


mjvezzani (approved):
There is nothing preventing this from going to prod. 

However, there _is_ some code that needs removing that somehow got missed when this went from local branch to development. The code that needs removing isn't getting called anywhere, so it _shouldn't_ have any impact.

The code that needs removing is:
- substituteAnalyticsReferences.sh

PR #1070, opened 2021-02-03
defect/DE8151 Fix hubspot form button styling

mjvezzani:
### Problem
On the crds-media homepage, the subscribe buttons that were part of the 2 different email subscription hubspot forms would occasionally be misstyled.

### Solution
According to this documentation on hubspot under the "Multiple form embed codes added to the same

tag" heading, it indicates that if you want to place multiple forms in the same containing div tag, you have to use the formInstanceId, suggesting that formInstanceId is not necessary when the forms are not within the same div tag. After removing formInstanceId from the various _includes files, 10/10 refreshes resulted in the buttons being styled correctly.

### Corresponding Branch
There are no corresponding branches associated with this work.

### Testing
Navigate to the crds media homepage. Refresh the page multiple times, each time confirming that the "subscribe" button is styled correctly.


kellgauthier (approved):

kellgauthier:
In theory, this should work, but I'm still seeing it misstyled both locally and on the deploy preview. 
Screen Shot 2021-02-03 at 12 33 49 PM

kellgauthier (approved):

PR #1073, opened 2021-02-09
Remove references to gtag() and switch off gtag substitution build pl…

mjvezzani:
…ugin

## Problem
Metarouter is showing a profusion of errors after gtag work was finished.

## Solution
We are manually rolling back the changes that were made to implement gtag.js


dholloran-ample (changes requested):
@mjvezzani, 2 things

1. Please remove the code instead of commenting it out.
1. From what I'm seeing there is still a gtag.js on the page. Where that is coming from I'm not sure. Are you seeing anything similar?

screenshot 2021-02-09 at 15 11 43

dholloran-ample:
@mjvezzani - I think `I think `<script async="" src="https://www.googletagmanager.com/gtag/js?id=UA-2977073-1"></script>` needs to be removed as well` still is in the page causing the duplicate tags maybe? I would think we should still remove it.

mjvezzani:
Needs to be a PR into master.

PR #1074, opened 2021-02-09
Hotfix/metarouter issues

mjvezzani:
## Problem
Proliferation in Metarouter errors requires we back out the gtag.js solution

## Solution
Manually remove code that triggers gtag.js


mjvezzani:
Still not right. Need to branch off of master for this fix.

PR #1084, opened 2021-02-15
DE8161/Resolves shows feedback comments

kellgauthier:
[Rally](https://rally1.rallydev.com/#/66096747656ud/custom/24109743995?detail=%2Fdefect%2F496977932712&fdp=true&previousTokens=%2F66096747656ud%2Fcustom%2F24109743995%3Fdetail%3D%252Fdefect%252F496977932712&ref=%2Fdefect%2F496977932712)
[Preview](https://deploy-preview-1084--crds-mediaint.netlify.app/shows/)

- [x] The video up top should auto-play. It is not obvious to me that it's a video at first with the play button being on the bottom left of the banner. _
- [x] I would imagine we would want a volume option _**Volume/Mute and Pause/Play button added**_
- [x] Under the collection name should be the episode name in the jumbotron (in the design it's "Episode 24: Why do we have a garage?") to make it more clear that it's the most recent episode that is playing. Right now we say Garage Bible Study twice in a row in the jumbotron. 
- [x] Font for this should be a bit smaller like in the design
- [x] The Play button and the More Videos button should be bigger like the design. They are more noticeable that way.
- [x] More Videos button should link to the collection page for whatever collection is featured. 
- [x] In the design when you scroll over a show image in the carousel that card increases in size
- [x] The collections cards in the Shows carousel should click to that collection landing page
- [x] Recently added should only be videos associated with a collection AND tagged as "feature on the shows landing page", right now I see student ministry collections videos but that is not a show. Content Administration issue with bad data on int.  _**This is a contentful / content management issue. Once the correct shows data is populated, it will display as anticipated in master. If warranted, I can delete what is currently in int for shows and recreate it to the design prior to merging this work into master**_
- [x] Recently added video thumbnails should be bigger like the design
- [x] The videos under recently added should be clickable.
- [x] The collection names in orange should be clickable to that collection landing page

Mobile 

- [x] The video is way cropped in on mobile vs the design
- [x] Play and More Videos buttons should be bigger like in the design
- [x] There is a fade to black from the video to the "Shows" Section in the design vs a clear line break


sarahshiffman:
Hey Kell, 

1. Can we move the episode title closer to the episode description?
2. The link for more videos will should go to /media/collections/[collection name] instead of using tags. We want them to fall on the collection landing page for each and not just a media page with tagged vidoes (i.e. /media/collections/garage-bible-study)
3. The shows cards still don't click anywhere? They should link to /media/collections/[collection name] like above!
4. The orange collection name links under Recently Added should /media/collections/[collection name] instead of using tags. 
5. We still need to see the collection description over the shows cards on mobile? Can we make it so that when you are on a square it animates and shows the description? Then when you scroll right the one to left scales back down without description text and the one on the screen scales up with description?

_includes/_shows-most-recent.html
```diff
@@ -1,10 +1,14 @@
 <div class="mobile-soft-left">
   <div class="recently-added-image-container">
+    <a href="{{ site.url }}/media/videos/{{ episode.slug }}">
```
dholloran-ample, line 3: I think you may be able to get the Episode's URL via `episode. permalink`. If you can let's use that instead of building up the URL.

_includes/_shows-most-recent.html
```diff
@@ -1,10 +1,14 @@
...
     <img class="recently-added-image" src="{{ episode.image.url | imgix: site.imgix }}?auto=format,compress&w=175" data-optimize-img>
+    </a> 
   </div>
   <div>
     {% for show in episode.collections %}
+    <a href="{{ site.url }}/media/collections/{{ show.slug }}">
```
dholloran-ample, line 9: I think you may be able to get the Shows' URL via `show. permalink`. If you can let's use that instead of building up the URL.

shows.html
```diff
@@ -6,25 +6,57 @@
...
+              {% for shows in featured_shows offset: 1  limit: 1 %}
               <p class="font-size-smaller soft-half-top text-white">
                 {{ shows.description }}
               </p>
-              {% endfor %}
-              <div class="shows-button-container soft-half-top">
-                <button class="btn video-play">Play</button>
-                <crds-button class="soft-half-left" href="//crossroads.net/media/videos" color="white" display="outline" text="More Videos"></crds-button>
+              <div class="video-control-container soft-half-top">
+                <a href="{{ site.url }}/media/collections{{ shows.slug }}"><button class="more-videos btn btn-white btn-outline">More Videos</button></a>
```
dholloran-ample, line 32: See the previous comment about `shows.permalink`

shows.html
```diff
@@ -38,13 +70,15 @@ <h2 class="font-family-condensed-extra text-uppercase text-white shows-landing-p
     <div class="row soft-bottom">
       <h2 class="font-family-condensed-extra text-uppercase flush-top text-white shows-landing-page-title mobile-soft-left">Shows</h2>
     </div>
-    <div class="row soft-bottom push-bottom">
+    <div class="row soft-bottom push-bottom carousel-shows">
```
dholloran-ample, line 73: See the previous comment about `shows.permalink`

dholloran-ample (changes requested):
@kellgauthier - Left a few inline comments about building up the URLs. Let me know if you have any questions.

_includes/_shows-most-recent.html
kellgauthier: `{{ show.permalink }}` did not result in the correct url. Same as the other suggestions.

kellgauthier:
Per direction from AG:

- add play button per design but title it "watch now" 
- remove play/pause button 
- keep mute/unmute button and make it smaller

kellgauthier:
Direction from client - move shows carousel up closer to the jumbotron

_assets/javascripts/components/shows-video-player.js
```diff
@@ -1,24 +1,39 @@
-let vid = document.getElementById("bgvid");
-let playButton = document.querySelector("#shows-jumbotron-content button");
+let vid = document.getElementById('bgvid');
+let muteButton = document.getElementById('mute-btn');
+let muteIcon = document.getElementById('volume-slash');
+let soundIcon = document.getElementById('volume-up');
+const hidden = 'hidden-icon';
+const display = 'icon-video';
```
tcmacdonald, line 6: I would recommend _not_ using ES6 keywords or features if you're not transpiling or using a module bundler. It's probably fine but you never know if somebody on IE11 will run into issues.

tcmacdonald:
Couple more issues to consider. Also, triple check the state of the carousel and alignment of certain items (mute icon, for example) on small screens. Its a little rough depending on the size of the viewport. 

Crossroads_Media

tcmacdonald (approved):

PR #1128, opened 2021-03-22
DE8222 Shows Card 404

silasbaur:
## Problem
Fix shows card orange link that is 404ing. I believe that part of the issue is that we have are serving the shows page through a 200 redirect to `https://crds-media.netlify.app/shows`. (this is a temporary solution for an issue that is on Netlify)

https://user-images.githubusercontent.com/32345656/112007907-6e7c5b00-8afb-11eb-9b20-0c8572590ad7.mov

## Solution
Not exactly sure what is going on but this link is for the collection detail page so I decided to grab the same url structure of this link that is being used on the carousel where it is not 404ing and hope that it fixes the issue

It is possible that maybe the `/collections/*` to `/media/collections/*` is not working because of the extra `/`. 

### Please note that this PR is point at master because of the redirect specified above


_includes/_shows-most-recent.html
```diff
@@ -6,7 +6,7 @@
   </div>
   <div>
     {% for show in episode.collections %}
-    <a href="{{ site.url }}/media/collections/{{ show.slug }}">
+    <a href="{{ site.url }}/media/collections{{ show.slug }}">
```
tcmacdonald, line 9: Hold up. This should evaluate whether `show.slug` has a preceding `/` character and only remove the literal character if that's the case.

tcmacdonald (changes requested):

_includes/_shows-most-recent.html
sarahshiffman: @silasbaur did you see this note above from Taylor?

PR #1131, opened 2021-03-24
US20858 teaching pastors

silasbaur:
## Task
Add section below the carousel called "Teaching Pastors" with 4 pastors displayed as in design below.

https://projects.invisionapp.com/d/main#/console/20964701/446645695/preview


dholloran-ample (changes requested):
@silasbaur - The picture for Chuck Mingo is cropped a little funny (see screenshot). In digging into this a little the image URLs seem like they are outputting the URL twice for some reason. It also has `456` as the width when it should be `228` since the image never gets larger than `228px`.

I've added both the current URL as an example and the URL structure that should fix it let me know if you have any questions.

## Current URL

`http://crds-media.imgix.net/63VRnIILnSF5JImGPBAhIc/e44e96984944aa080b09f8a32e7ef13b/Mingo_Chuck.jpg?http://crds-media.imgix.net/63VRnIILnSF5JImGPBAhIc/e44e96984944aa080b09f8a32e7ef13b/Mingo_Chuck.jpg=undefined&w=456`



##I think this URL will solve the issue:

`http://crds-media.imgix.net/63VRnIILnSF5JImGPBAhIc/e44e96984944aa080b09f8a32e7ef13b/Mingo_Chuck.jpg?w=228&h=228&crop=bottom,center&fit=crop`



## Screenshot

screenshot 2021-03-25 at 10 07 41

dholloran-ample (approved):

PR #1149, opened 2021-04-12
refactors shows

kellgauthier:
## Problem
The way that /shows was set up initially was incorrect and causing concern with future collection work (i.e. Humans of Crossroads).

[Story on Rally ](https://rally1.rallydev.com/#/66096747656ud/custom/24109743995?detail=%2Fuserstory%2F515536349964)
## Solution
The data for shows has been restructured within the int and demo environments within Contentful, and the ruby generator has been minimized to only pull in episodes by video associated with a collection. 

## Testing
Once this approach is validated, we will need a crds-contentful-migration to ensure the master environment within contentful matches what is currently in int and demo.


dholloran-ample (commented):
@kellgauthier - I re-worked the show episodes generator to make it so only videos that were part of the Shows collection are added. The previous implementation adds any video with a collection.

Currently, there is only one item that I saw on int "Garage Bible Study". Could you add a few more and test that what I changed works for you? Once that's complete I think we can merge.

PR #1150, opened 2021-04-13
DE8027/shows-background-video-error

kellgauthier:
## Problem
The ad detection infrastructure is called Ad Tagging. Ad Tagging works by matching resource requests against a filter list (see how the list is generated) to determine if they’re ad requests. Any requests matching the filter are tagged as ads. Further, requests (and some DOM elements such as iframes) made on behalf of previously tagged scripts are also tagged as ads by the AdTracker. An iframe will be marked as an ad iframe if its url matches the filter list, if tagged script is involved in the creation of the iframe, or if its parent frame is an ad iframe. The main frame of a page will never be tagged as an ad. Any request made within an ad iframe is considered an ad resource request.

## Solution
Instead of having javascript create and load an <iframe> through the YouTube API - this injects a youtube id into an embed iframe that's already loaded when the page loads, and show allow for the video to play without being blocked by Chrome's Ad Tagger.


dholloran-ample (changes requested):
@kellgauthier - This is not working for me. 

I have a theory that the `<iframe />` does not exist on the page when this code runs. I'm pretty sure Media has jQuery available so you could **maybe** do the following. I have not tested it this is just a guess.

```js
$(document).ready(function () {
    // All the code in the background-video.js file
})
```

PR #1154, opened 2021-04-15
US20829 humans of crossroads index page

kellgauthier:
## Task
Create index page based off of design below and create a generator to pull in the correct articles associated with the Humans of Crossroads collection.
[Design](https://invis.io/JC10HVK0Z25B#/447569898_crds-Humans-Of-Crossroads-02) 
[Rally](https://rally1.rallydev.com/#/66096747656ud/custom/24109743995?detail=%2Fuserstory%2F514488428160)


_plugins/generators/humans_article_generator.rb
```diff
@@ -0,0 +1,12 @@
+module Jekyll
```
dholloran-ample, line 1: I may be wrong but I don't think this Generator is needed. It looks to me that it is only transforming `site.collections['humans_of_crossroads']` so it can be accessed as `site.data['humans_of_crossroads']`. 

If Jekyll is already loading the `site.collections['humans_of_crossroads']` automatically then let's just use that instead. I think you would access it as `site.humans_of_crossroads` in Liquid but not 100% sure.

dholloran-ample (changes requested):
@kellgauthier - Left one inline comment about the Generator

dholloran-ample (dismissed):

dholloran-ample (changes requested):
Adding these for reference from our Slack convo @kellgauthier.

1. Adjust container width
2. We don't need to worry about the missing `<hr />` for now
3. Adjust spacing vertically between images
4. Fix images breaking out of their container on smaller screens

screenshot-1
screenshot-2

dholloran-ample (approved):

PR #1157, opened 2021-04-21
Hotfix humans generator

tcmacdonald:
## Problem
Data populating the HOC section on the site is not working correctly. 

## Solution
This PR removes the generator and replaces it with a `preinit` hook to populate a new HOC collection at the start of each build. 

## Testing
https://607f914f3f5df23bf6c9aef7--crds-mediaint.netlify.app/humans-of-crossroads/


silasbaur (changes requested):
Looks good! I have two questions:
1. Why did you delete the humans article template?
2. The humans articles still show up in article directory, which then makes them show up on the article index page. So currently we have the humans article in two different places. In your hook can we remove the humans articles out the article directory?

silasbaur (approved):

